PR #22, opened 2023-08-04
Update hugging face notebook

TattdCodeMonkey:
add Colab link to hugging face and remove outputs


notebooks/integrations/hugging-face/loading-model-from-hugging-face.ipynb
```diff
@@ -334,40 +197,29 @@
...
-    }
    ]
   },
   {
    "cell_type": "markdown",
+   "metadata": {
+    "id": "IW-GIlH2OxB4"
+   },
    "source": [
     "### Create Index and mappings for Blogs\n",
```
joemcelroy, line 208: suggestion: condense the field setup to just body_content, title, url & id, as we dont use these fields for the search and simplifies for readability.

notebooks/integrations/hugging-face/loading-model-from-hugging-face.ipynb
```diff
@@ -26,120 +19,38 @@
     "\n",
     "### Prerequisities\n",
     "Before you start make sure you have Elasticsearch cluster running. The cluster must have at least one machine learning (ML) node with enough (4GB) memory."
```
joemcelroy, line 21: nitpick: Do we need 4gb memory? Could we get away with a smaller node with this being a smaller model?

joemcelroy:
im not keen now on how we are loading blog content, as we are embedding on a very dense field that contains html. For simplicity, i think we should use the sample dataset that some of the [examples used](https://github.com/elastic/elasticsearch-labs/blob/c91e05ee2dcc3fb8005d7b090ce5dce6874b7e98/colab-notebooks-examples/search/data.json) just so we avoid topics like chunking, html stripping. We can then provide a nice response example.

notebooks/integrations/hugging-face/loading-model-from-hugging-face.ipynb
radoondas: You are right. The model used here is only 416MB. We can go with 2GB node.

radoondas:
> 

I do not mind changing the dataset, but the results will be mixed feelings. 
The search with NLP works better with a bigger dataset. But, if the book is only for how to load the NLP model from hugging face, then OK.

joemcelroy:
closing this now that saarika has merged her eland improvements.

PR #35, opened 2023-08-09
Add chatbot notebook

demjened:
TBD

### DIFF
--- a/notebooks/generative-ai/chatbot.ipynb
+++ b/notebooks/generative-ai/chatbot.ipynb
@@ -1,0 +1,529 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Chatbot with LangChain conversational chain and OpenAI 🤖💬\n",
+    "\n",
+    "In this notebook we'll build a chatbot that can respond to questions about custom data, such as policies of an employer.\n",
+    "\n",
+    "The chatbot uses LangChain's `ConversationalRetrievalChain` and has the following capabilities:\n",
+    "- Answer questions asked in natural language\n",
+    "- Run hybrid search in Elasticsearch to find documents that answer the question\n",
+    "- Extract and summarize the answer using OpenAI LLM\n",
+    "- Maintain conversational memory for follow-up questions"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Requirements 🧰\n",
+    "For this example, you will need:\n",
+    "\n",
+    "- Python 3.6 or later\n",
+    "- An Elastic deployment with minimum **2GB machine learning node**\n",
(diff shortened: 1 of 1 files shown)


joemcelroy:
so good! Only comment is re-check the requirements - im pretty sure we dont need an ML node for this as inference is done via langchain llm (openai)

joemcelroy (approved):

notebooks/generative-ai/chatbot.ipynb
```diff
@@ -0,0 +1,433 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "vscode": {
+     "languageId": "shellscript"
+    }
+   },
+   "outputs": [],
+   "source": [
+    "%pip install -U langchain==0.0.245 jq openai elasticsearch tiktoken"
```
joshdevins, line 59: It looks like `jq` and `toktoken` aren't used. I think they can be removed?

I'd also be inclined to either not pin `langchain`'s version, or provide something like `>=` if there is a minimum version requirement.

joshdevins (changes requested):
LGTM! Nice and simple. I like it.

We ought to start to standardise on our quoting in this repo as well. The age-old debate about single, double and triple quotes in Python 🙄. We just need to pick a practice and make it uniform, but I've seen multiple styles in the repo so far. Not a blocker for this PR though!

cc @joemcelroy

notebooks/generative-ai/chatbot.ipynb
joemcelroy: > I'd also be inclined to either not pin langchain's version, or provide something like >= if there is a minimum version requirement.

Unfortunately theres a bug with elasticsearch langchain vectorstore that was introduced recently. My changes (soon to be merged!) will update this code and no longer need to peg the version.

notebooks/generative-ai/chatbot.ipynb
```diff
@@ -0,0 +1,433 @@
...
+    "workplace_docs = json.loads(response.read())\n",
+    "\n",
+    "print(f'Successfully loaded {len(workplace_docs)} documents')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As we're chatting with our bot, it will run semantic searches on the index to find the relevant documents. In order for this to be accurate, we need to split the full documents into small chunks (also called passages). This way the semantic search will find the passage within a document that most likely answers our question.\n",
```
leemthompo, line 191: nit: I'd suggest making the _splitting/chunking_ step its own sub-heading, so you'd have two subheadings something like:

- Load ~and process~ documents
- Chunk documents into passages

notebooks/generative-ai/chatbot.ipynb
demjened: Removed `jq` and `tiktoken`.

PR #40, opened 2023-08-11
OpenAI RAG Streamlit notebook example

lio-p:
Add notebook to show how to add a simple web interface to explore an example of a Generative AI Application using Elasticsearch and OpenAI


joemcelroy:
i think simplify the notebook so focuses on streamlit + elasticsearch + openai. show only an example of inserting documents in and then have the streamlit example. That should shorten the notebook and make it easier to follow .

Link them to the openai notebook that we have if they want more details.

lio-p:
@joemcelroy Thanks for the review. I simplified the notebook

joemcelroy:
@lio-p looks good. Couple of minor changes:
- could you remove the output from frames where they do not add value? Example of this is the install packages. https://github.com/elastic/elasticsearch-labs/blob/b739eefb5af5e2f19d614d40776a4e0ae18da277/supporting-blog-content/openai-rag-streamlit/openai_rag_streamlit.ipynb
- Suggestion: add an image on the top with an example of the streamlit UI? just so they can visualise what they will get.

joemcelroy (approved):

PR #55, opened 2023-08-29
Fix incorrect file url

wonseop:
In the "Semantic Search Quick Start" example,
404 error occurs because the url of the index test data is incorrect.
I modified the url correctly.


joemcelroy (approved):

joemcelroy:
thank you!

PR #65, opened 2023-09-14
Update 03-ELSER.ipynb: Fix wrong lookup

herrBez:
Fix lookup of field that failed because https://raw.githubusercontent.com/elastic/elasticsearch-labs/main/notebooks/search/movies.json does not contain `text` but only the `field` plot


joemcelroy (approved):

joemcelroy:
thank you!

PR #69, opened 2023-09-25
Add example query rules notebook

kderusso:
Adds an example query rules notebook using the examples from the blog: https://www.elastic.co/guide/en/elasticsearch/reference/master/search-using-query-rules.html


notebooks/search/05-query-rules.ipynb
```diff
@@ -0,0 +1,437 @@
...
+    {
+      "cell_type": "markdown",
+      "id": "87773ce7",
+      "metadata": {
+        "id": "87773ce7"
+      },
+      "source": [
+        "# Search using query rules\n",
+        "\n",
+        "<a target=\"_blank\" href=\"https://colab.research.google.com/github/kderusso/elasticsearch-labs/blob/kderusso/query-rules-notebook/notebooks/search/05-query-rules.ipynb\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>\n",
```
kderusso, line 12: Note: I will update these links before merging, but the link to my repo is required to run in colab.

notebooks/search/05-query-rules.ipynb
```diff
@@ -0,0 +1,437 @@
...
+      "cell_type": "code",
+      "execution_count": null,
+      "id": "ffc5fa6f",
+      "metadata": {
+        "id": "ffc5fa6f",
+        "scrolled": false
+      },
+      "outputs": [],
+      "source": [
+        "!pip install -qU elasticsearch sentence-transformers==2.2.2"
```
joemcelroy, line 58: nitpick: dont need sentence transformers here

kderusso: Thanks for the catch! You caught me on copy-pasta 😂

notebooks/search/05-query-rules.ipynb
```diff
@@ -0,0 +1,437 @@
...
+        },
+        "id": "25c618eb",
+        "outputId": "30a6ba5b-5109-4457-ddfe-5633a077ca9b"
+      },
+      "outputs": [
+        {
+          "name": "stdout",
+          "output_type": "stream",
+          "text": [
+            "{'name': 'instance-0000000000', 'cluster_name': '1a56ad21587c44d3930932eb9fa1d8e8', 'cluster_uuid': 'gX4zlwtlR4qhZpp1SPm4Yg', 'version': {'number': '8.8.2', 'build_flavor': 'default', 'build_type': 'docker', 'build_hash': '98e1271edf932a480e4262a471281f1ee295ce6b', 'build_date': '2023-06-26T05:16:16.196344851Z', 'build_snapshot': False, 'lucene_version': '9.6.0', 'minimum_wire_compatibility_version': '7.17.0', 'minimum_index_compatibility_version': '7.0.0'}, 'tagline': 'You Know, for Search'}\n"
```
pquentin, line 144: This suggests that Elasticsearch 8.8 is enough to use query rules. Should we say somewhere that this was introduced in Elasticsearch 8.10?

carlosdelest (approved):
I've tested it and LGTM!

One small suggestion - I think it would be useful to compare the results of the rule being applied with the rule not being applied. Users could compare the results from the query and check the rule effects.

notebooks/search/05-query-rules.ipynb
```diff
@@ -0,0 +1,593 @@
...
+      "outputs": [],
+      "source": [
+        "response = client.search(index=\"products_index\", body={\n",
+        "    \"query\": {\n",
+        "      \"multi_match\": {\n",
+        "          \"query\": \"reliable wireless charger for iPhone\",\n",
+        "          \"fields\": [ \"name^5\", \"description\" ]\n",
+        "      }\n",
+        "    }\n",
+        "})\n",
```
pquentin, line 270: I have a small nit here. The Elasticsearch Python client, starting with version 8.0, supports parameters for the first level of body keys. Which means you can write:

```python3
response = client.search(index="products_index", query={"multi_match": ...})
```

The `body` form you're using here is fine (it was deprecated in 8.0 but that deprecation will go away in 8.11 or maybe earlier) but I think it would be nice to be consistent throughout the notebook. So I would either switch to `body` for `client.query_ruleset.put` or switch to `query` for `client.search`. I think it would make the `query_ruleset` call feel less magic.

pquentin (approved):
Thanks! LGTM. I like how this is short and to the point but still using real data, making this is a great starting point to use query rules.

notebooks/search/05-query-rules.ipynb
kderusso: Done, thanks!

PR #76, opened 2023-10-12
notebook fixes

miguelgrinberg:
This is my initial pass at the search notebooks, with an eye towards consistency. Unfortunately it is impossible to review diffs of notebooks, so below is a quick list of the changes that I've made:

### 00-quick-start.ipynb
- Removed pin on `sentence-transformers`
- Removed `device` argument when initializing the sentence transformer, as it is optional and the default is appropriate.
- Moved the index delete that was at the bottom to the section in which the index is created.
- Removed all uses of the deprecated `body` argument.
- Print a "no results" message in the `pretty_response()` function.
- Fixed issues with the `term` filter and a hyphenated text field.

### 01-keyword-querying-filtering.ipynb
- Removed unused dependencies
- Added a note indicating that this notebook uses the index created in the quick start.
- Clear separation between query context searches vs. filter context searches.
- Many added examples.

### 02-hybrid-search.ipynb
- Remove instructions to create the index and refer people to the quickstart instead.
- Print a "no results" message in the `pretty_response()` function.

### 03-ELSER.ipynb
- No changes, since this is actively being edited in another PR.

### 04-multilingual.ipynb
- Removed all uses of the deprecated `body` argument.
- Removed dataset from the notebook and relocated it to a `articles.json` file. (I left a temporary URL to help with reviews, will remove before merging)
- Print a "no results" message in the `pretty_response()` function.
- Minor reorganization of the search section.

## 05-query-rules.ipynb
- Print a "no results" message in the `pretty_response()` function.


joemcelroy:
## 00-quick-start.ipynb
- suggestion: simplify the filtering section, remove the advanced filtering example
- suggestion: remove the hybrid search section, thats cover in the hybrid search notebook
- suggestion: combine the filtering and example keyword filtering example into one. lean on the querying filtering notebookto introduce more examples.

joemcelroy:
## 01-keyword-querying-filtering.ipynb
- Currently this uses the dataframe example. Update it to use the `pretty_response_transform` example
- filtering should include an example of using a term query

miguelgrinberg:
Okay! Made the simplifications requested. Let me know what you think.

joemcelroy:
## 00-quick-start
One thing i missed was the filtering example should use a keyword type subfield, which is accessible via `publisher.keyword`. We are using the term query so should be targeted on the raw unanalyzed value, which is in the keyword field. `publisher` is a text field and gone through an analyzer. This means if you use the text field, you may include documents which contain Addison. Using a keyword, it must match exactly.

miguelgrinberg:
Okay, a few more changes on notebooks 00 and 01 to always use keyword fields in term filters.

joemcelroy:
looks good!

Only last thought is with 01-filters - we introduce straight into boolean clause and dont explain the common filter activities like term, range (numeric + date) and jump into talking about combining multiple filters. Should we add additional context here?

joemcelroy (approved):

PR #78, opened 2023-10-12
ELSER: Use API to load and deploy ELSER model

joemcelroy:
instead of using the UI to load and deploy the model, rely on the API calls to do this, so the activity is fully done within the notebook

### DIFF
--- a/notebooks/search/03-ELSER.ipynb
+++ b/notebooks/search/03-ELSER.ipynb
@@ -19,40 +19,6 @@
    "attachments": {},
    "cell_type": "markdown",
    "metadata": {
-    "id": "Y01AXpELkygt"
-   },
-   "source": [
-    "# 🧰 Requirements\n",
-    "\n",
-    "For this example, you will need:\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


miguelgrinberg (approved):
Everything works, so :+1: 
But I do think it would be good to be nice to readers and let them know how to remove anything that can cost them money if left running. Now that we aren't using the UI to deploy the model nobody is going to figure out where to go to destroy stuff.

joemcelroy:
> Note that I did get the "not ready to be deployed". If we can't make the download block maybe we can add a loop that waits for the ready flag before moving on? This is so that the notebook can run through everything unattended.

Yeh this one a tough one to call. I kept it light for now but lets see what feedback is like and we can add more logic to this. Im apprehensive increasing complexity with inline code. Maybe its something we could consider if we have a library of "helpers" that we can manage separately from the notebooks.  

> But I do think it would be good to be nice to readers and let them know how to remove anything that can cost them money if left running. Now that we aren't using the UI to deploy the model nobody is going to figure out where to go to destroy stuff.

Good suggestion. Added a small sentence to remind developers to scale down.

PR #84, opened 2023-10-16
[Notebook] Update chatbot notebook to use Langchain Memory Store

joemcelroy:
Now that our PR has been merged in langchain, use memory store thats part of the langchain lib instead.


miguelgrinberg (approved):
I'm not set up for OpenAI so I cannot run this code, but the changes are minimal and look good.

PR #93, opened 2023-10-18
first commit to start this issue https://github.com/salgado/music-sea…

salgado:
…rch/issues/1#issuecomment-1766366347 and https://github.com/elastic/search-labs-elastic-co/issues/93#issuecomment-1765153661


serenachou:
@joemcelroy  can you review this PR please?

supporting-blog-content/music-search/elastic_music_search.ipynb
joemcelroy, line 1: Looks good. Only suggestion I can make reading the code is to make use of the bulk helper to add in one call, rather than indexing each item individually. https://github.com/elastic/elasticsearch-labs/blob/main/notebooks/search/00-quick-start.ipynb

Also to mention that after the documents have been added, it would be good to have a refresh flag so documents are immediately searchable. If the user executes all the commands at once, there might be a chance where it hasn't refreshed and no result will be returned.

PR #100, opened 2023-10-29
adding code for Elasticsearch as a GenAI Caching Layer Blog

jeffvestal:
contet for forthcoming blog 

Elasticsearch LLM Cache Library
==================================
This library provides an Elasticsearch-based caching mechanism for Language Model (LLM) responses.
Through the ElasticsearchLLMCache class, it facilitates the creation, querying, and updating
of a cache index to store and retrieve LLM responses based on user prompts.

Key Features:
-------------
- Initialize a cache index with specified or default settings.
- Create the cache index with specified mappings if it does not already exist.
- Query the cache for similar prompts using a k-NN (k-Nearest Neighbors) search.
- Update the 'last_hit_date' field of a document when a cache hit occurs.
- Generate a vector for a given prompt using Elasticsearch's text embedding.
- Add new documents (prompts and responses) to the cache.

### DIFF
--- a/supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
+++ b/supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
@@ -1,0 +1,220 @@
+"""
+Elasticsearch LLM Cache Library
+==================================
+This library provides an Elasticsearch-based caching mechanism for Language Model (LLM) responses.
+Through the ElasticsearchLLMCache class, it facilitates the creation, querying, and updating
+of a cache index to store and retrieve LLM responses based on user prompts.
+
+Key Features:
+-------------
+- Initialize a cache index with specified or default settings.
+- Create the cache index with specified mappings if it does not already exist.
+- Query the cache for similar prompts using a k-NN (k-Nearest Neighbors) search.
+- Update the 'last_hit_date' field of a document when a cache hit occurs.
+- Generate a vector for a given prompt using Elasticsearch's text embedding.
+- Add new documents (prompts and responses) to the cache.
+
+Requirements:
+-------------
+- Elasticsearch
+- Python 3.6+
+- elasticsearch-py library
+
+Usage Example:
+--------------
+```python
+from elasticsearch import Elasticsearch
+from elasticsearch_llm_cache import ElasticsearchLLMCache
+
+# Initialize Elasticsearch client
+es_client = Elasticsearch()
+
+# Initialize the ElasticsearchLLMCache instance
+llm_cache = ElasticsearchLLMCache(es_client)
+
+# Query the cache
+prompt_text = "What is the capital of France?"
+query_result = llm_cache.query(prompt_text)
+
+# Add to cache
+prompt = "What is the capital of France?"
+response = "Paris"
+add_result = llm_cache.add(prompt, response)
+```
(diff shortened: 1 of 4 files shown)


supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
```diff
@@ -0,0 +1,220 @@
...
+
+        :param es_client: Elasticsearch client object.
+        :param index_name: Optional name for the index; defaults to 'llm_cache'.
+        :param es_model_id: Model ID for text embedding; defaults to 'sentence-transformers__all-distilroberta-v1'.
+        :param create_index: Boolean to determine whether to create a new index; defaults to True.
+        """
+        self.es = es_client
+        self.index_name = index_name or 'llm_cache'
+        self.es_model_id = es_model_id
+        if create_index:
```
joemcelroy, line 82: suggestion: dim size as an optional arg, default value of 768 and pass in to the create_index.

supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
```diff
@@ -0,0 +1,220 @@
...
+from typing import Dict, List, Optional
+from elasticsearch import Elasticsearch
+import logging
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+class ElasticsearchLLMCache:
```
joemcelroy, line 64: suggestion: we should have a threshold for when an entry is considered as stale, based on the `create_date`. This should be taken into consideration.

for future: we could add a stale-while-revalidate concept where we return stale response immediately and perform a query async to re-update refresh cache entry.

supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
```diff
@@ -0,0 +1,220 @@
...
+        doc = {
+            "prompt": prompt,
+            "response": response,
+            "create_date": datetime.now(),
+            "last_hit_date": datetime.now(),
+            "prompt_vector": prompt_vector,
+            "source": source  # Optional
+        }
+        self.es.index(index=self.index_name, document=doc)
+        return {'success caching new prompt & response': True}
```
joemcelroy, line 220: suggestion: Simplify this return?

joemcelroy (approved):
A couple suggestions, nothing blocking IMO. Nice work!

supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
jeffvestal: @joemcelroy I have that on line 86 for the `create_index` method. 
I left it out of init since you don't need it if you aren't going to create an index. But If you think it makes more sense in `__init__` I can move it there?

supporting-blog-content/elasticsearch_llm_cache/elasticsearch_llm_cache.py
jeffvestal: > stale-while-revalidate 
I like that idea
I added a  default of 1 year filter range, that can be changed

PR #105, opened 2023-11-03
Rag chatbot langchain

joemcelroy:
This updates the langchain notebook by:

- stepping through the changes from simple to advanced
- Update the documents to use JSONLoader
- moved package imports to where the code is rather than at the top
- cleaned up the headings
- updated to langchain "LCEL"

One slight change is this notebook now persists a file (the JSON data file) onto disk before it loads it into JSONLoader. I've tested this on google colab and works, just something to be aware of. 

I've also removed the empty embeddings notebook because its unlikely we will do this.


notebooks/generative-ai/question-answering.ipynb
```diff
@@ -88,34 +90,44 @@
...
+        "import json\n",
+        "\n",
         "url = \"https://raw.githubusercontent.com/elastic/elasticsearch-labs/main/example-apps/workplace-search/data/data.json\"\n",
         "\n",
         "response = urlopen(url)\n",
+        "data = json.load(response)\n",
         "\n",
-        "workplace_docs = json.loads(response.read())\n"
+        "with open('data.json', 'w') as json_file:\n",
+        "    json.dump(data, json_file)\n"
```
miguelgrinberg, line 130: Seems strange that you are downloading this file, just to save it again, decode and re-encoding the JSON. Doesn't look like there is a JSON URL downloader in langchain though, so maybe this is the best that can be done.

notebooks/generative-ai/question-answering.ipynb
```diff
@@ -172,12 +203,17 @@
...
       "outputs": [],
       "source": [
         "documents = vector_store.from_documents(\n",
-        "    docs, embeddings, es_cloud_id=CLOUD_ID, es_user=CLOUD_USERNAME, es_password=CLOUD_PASSWORD, index_name=\"workplace_index\"\n",
+        "    docs, \n",
+        "    embeddings,  \n",
+        "    index_name=\"workplace_index\",\n",
+        "    es_cloud_id=CLOUD_ID,\n",
+        "    es_user=CLOUD_USERNAME,\n",
+        "    es_password=CLOUD_PASSWORD\n",
```
miguelgrinberg, line 216: This is only appropriate for a relatively short list of documents, right? You would never take this approach for indexing a large collection of docs. Should we show how to work with a large dataset, either instead of in addition to this short dataset example? Or at least mention that this is not a scalable solution?

notebooks/generative-ai/question-answering.ipynb
```diff
@@ -208,43 +244,375 @@
...
+        "retriever = vector_store.as_retriever()\n",
+        "\n",
+        "llm = OpenAI(openai_api_key=OPENAI_API_KEY)\n",
+        "\n",
+        "ANSWER_PROMPT = ChatPromptTemplate.from_template(\n",
+        "\"\"\"\n",
+        "Use the following passages to answer the user's question.\n",
+        "Each passage has a SOURCE which is the title of the document. When answering, cite source name of the passages you are answering from below the answer in a unique bullet point list.\n",
+        "\n",
+        "If you don't know the answer, just say that you don't know, don't try to make up an answer.\n",
```
miguelgrinberg, line 324: In your experience, does this help reduce hallucinations? Isn't it a bit like asking a drunk person to act sober?

notebooks/generative-ai/question-answering.ipynb
```diff
@@ -208,43 +244,375 @@
...
+        "    standalone_question | retrieved_documents | answer\n",
+        ")\n",
+        "\n",
+        "ans = chain.invoke({ \n",
+        "  \"question\": \"What are their objectives?\", \n",
+        "  \"chat_history\": [\n",
+        "    \"What is the nasa sales team?\",\n",
+        "    \"The sales team of NASA consists of Laura Martinez, the Area \"\n",
+        "    \"Vice-President of North America, and Gary Johnson, the Area \"\n",
+        "    \"Vice-President of South America. (Sales Organization Overview)\",\n",
```
miguelgrinberg, line 602: Should this say `(SOURCE: Sales Organization Overview)` to be consistent with the responses? Or else remove the `SOURCE:` prefixes in the example output?

Note that I'm not running the code, just looking at the logic and the outputs that you saved.

notebooks/generative-ai/question-answering.ipynb
```diff
@@ -208,43 +244,375 @@
...
+        "  ]\n",
+        "})\n",
+        "\n",
+        "print(\"---- Answer ----\")\n",
+        "print(ans[\"answer\"])"
```
miguelgrinberg, line 607: Maybe we want to show how the answer is added to the chat history so that it is available for the next question?

notebooks/generative-ai/question-answering.ipynb
joemcelroy: Yeh its not great but unfortunately it doesn't support loading from url or from a blob so its the best I can do. I prefer doing it via loader than manually as its easier to read and relate to their docs.

notebooks/generative-ai/question-answering.ipynb
joemcelroy: Its also fine for ingesting large amount of docs, which developers use for this. Behind the scenes it uses bulk helper which chunks requests.

notebooks/generative-ai/question-answering.ipynb
joemcelroy: Yes, you need to instruct this otherwise the LLM will try and use its own knowledge to answer the question, if its not in the context and likely make up an answer.

notebooks/generative-ai/question-answering.ipynb
miguelgrinberg: Okay, so then I don't understand how this works. What does the `docs` local variable holds? It seems it has the entire database of chunked documents. My thinking was that this is only viable when your entire database fits in memory, which I think is probably not the case for a lot of people.

notebooks/generative-ai/question-answering.ipynb
miguelgrinberg: Then wouldn't you say somthing like "if the question isn't answered in the context information, say I don't know; don't try to make up an answer from other sources".

notebooks/generative-ai/question-answering.ipynb
joemcelroy: I prefer the readability of just inlining the answer TBH.

notebooks/generative-ai/question-answering.ipynb
joemcelroy: good spot, updated.

notebooks/generative-ai/question-answering.ipynb
joemcelroy: Yeh - being terse as smaller the prompt, less cost to generate. Happy to discuss later on optimizing the prompt. IMO it will be something i would expect the developer to tweak.

notebooks/generative-ai/question-answering.ipynb
joemcelroy: discussed offline.

notebooks/generative-ai/question-answering.ipynb
miguelgrinberg: My issue with this prompt is that it implies that the AI knows what it knows and what it doesn't, and is able to make decisions based on that, which is false. All the AI can do is find sequences of words that are probabilistically more likely to agree with the prompt based on what it has seen during training. "If you don't know the answer..." is very vague in that respect, something like "If the answer isn't in the context provided..." is in my opinion a much better guidance.

miguelgrinberg (approved):
I have added some more context regarding my issues with the prompt that you are using, but we can chat about prompts later if you like.

PR #106, opened 2023-11-03
[Internal Knowledge Search] Add v1

timgrein:
This is the first version of the workplace search reference app, which provides a very basic search functionality using the `search-application-client`, Search Applications and content indices powered by connectors.

Known issues, which will be addressed in an iterative manner:
- Use facets for data source filtering (in general use facets)
- Autocompletion
- Query history
- Lift mappings to the UI
- Introduce an intermediate backend responsible for handling the metadata for the application (mappings etc.)
- Pagination
- More sorting capabilities
- ...


example-apps/internal-knowledge-search/app-ui/.env
```diff
@@ -0,0 +1,3 @@
+SEARCH_APP_NAME=some-search-application
```
yansavitski, line 1: Better to rename in .env.example as common practise for this kind of file

example-apps/internal-knowledge-search/app-ui/src/pages/SearchPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+    //TODO: simplify query state and move it to one place
+    const [query, setQuery] = useState<string>("");
+
+    const {appName, apiKey, searchEndpoint} = useSelector((state: RootState) => state.searchApplicationSettings);
+    const {sorts} = useSelector((state: RootState) => state.sort);
+    const indexFilter = useSelector((state: RootState) => state.filter)["filters"]["Data sources"].values;
+
+
+    useEffect(() => {
+        const fetchData = async () => await handleSearchSubmit();
```
yansavitski, line 47: No big deal, but you can simply call handleSearchSubmit() is doesn't make sense to wrap it and call

example-apps/internal-knowledge-search/app-ui/src/pages/SearchPage.tsx
```diff
@@ -0,0 +1,111 @@
...
+        fetchData();
+    }, [indexFilter, sorts, appName, apiKey, searchEndpoint]);
+
+    const handleSearchSubmit = async () => {
+        try {
+            setLoading(true);
+
+            const client = SearchApplicationClient(appName, searchEndpoint, apiKey, {
```
yansavitski, line 55: As a future improvement u can initiate client when appName or searchEndpoint or apiKey change and here just call client()...
Every call of client initiates new query builder chain so it redundant to initiate new client every time on request

yansavitski (approved):
LGTM
There is some comments, but they are not blocker

example-apps/internal-knowledge-search/app-ui/.env
timgrein: Will add this as an improvement issue for later, thanks!

example-apps/internal-knowledge-search/app-ui/src/pages/SearchPage.tsx
timgrein: I'm actually using this to workaround the built-in caching of the client as there's no way to disable the caching through the public API, right? So for demo purposes it's better to always show the actual newest results for a query instead of having to waiting for the internal caching time. Maybe we could expose these settings in the `search-application-client` through its public API?

PR #127, opened 2023-11-29
Add a book dataset with chunked synopsis

lio-p:
Add a book dataset with chunked synopsis. Can be used for different examples. 

It will be first used as part of a [post](https://discuss.elastic.co/t/348129) for the [Advent calendar posts](https://github.com/elastic/community/issues/1424).


miguelgrinberg (approved):
Is this referenced (or going to be) in a blog article? Because without any references to it people are unlikely to find it.

PR #132, opened 2023-12-05
Updated chatbot gif

miguelgrinberg:
My simple intention of updating the gif snowballed into a number of issues/fixes, all included in this PR:

- Use temperature=0 for all LLMs
- The frontend inserted spurious `\n` in the LLM response stream
- The RAG prompt was adjusted to be more clear and specific with regards to how the sources should be given in the response


example-apps/chatbot-rag-app/frontend/src/store/provider.tsx
```diff
@@ -233,7 +233,7 @@ export const thunkActions = {
         
               dispatch(actions.setStatus({ status: AppStatus.Done }))
             } else {
-              message += message && event.data.trim() === '' ? '\n' : event.data
```
miguelgrinberg, line 236: The issue here is that many times the stream returned by the LLM is just a space between words. That should be preserved literally in the frontend, I think.

joemcelroy (approved):

PR #133, opened 2023-12-06
Parent child retriever examples

joemcelroy:
This PR demonstrates:
- Example 1: Using nested dense vector for document -> Full document + nested multiple passages.
- Example 2: Using nested dense vector for document -> chunked document + nested smaller passages.

Fixed couple issues on:
- remove unneeded function in bm25 example
- remove shard settings so can run on serverless


miguelgrinberg:
Comments:
- Not including the commands to upload the model to the ML node means that most people will not be able to run this notebook to completion.
- In 8.11 you do not need to specify the vector dimensions or the `index` variable. Actually I think with 8.11 you can get away w/o declaring any properties on the nested field.
- At some point you had reservations with using dot product for the minilm-l6-v2 model. I think it is fine based on the information [here](https://www.sbert.net/docs/pretrained_models.html), but I mention it just in case you prefer to switch to cosine.
- I found the big and small chunking difficult to understand. In reality, when you create the larger chunks you do not preserve a link to the complete document, so these large chunks are effectively parent documents in their own right. So I'm not sure doing a two-level splitting contributes much to the example. Both examples could use complete doc and nested passages, with example 1 showing how to get only the passages, and example 2 showing how to get the parent and the passages. No need to change anything, just thinking out loud here that I think I would have found it easier to follow without the large chunks...

joemcelroy:
Hey @miguelgrinberg 

Going to do the following updates:
- add the part on how to load the ML model into the notebook so can be run to completion
- update not declaring the nested field

PR #141, opened 2023-12-18
add spo dls to internal knowledge search

seanstory:
### Description
The goals of this PR are:
- to introduce Sharepoint Online as a searchable content type
- to provide an example of using document level security in a search experience

To that end, several significant changes were made.
1. The configuration of a single API key was replaced with configuration for a username/password. This is because API keys cannot be used to create other API keys with different role descriptors.
2. A new configuration of a "search persona" was added. This defaults to a hardcoded string of "admin", but if the configured search application has an index with DLS enabled, the first 30 identities will be fetched from the identities index and will be populated as options in the dropdown.
3. The SearchApplicationSettings model was augmented to include these new settings

### Requested Review
React is not my wheelhouse, and I know the diff is not currently well-organized. Please feel free to provide "obvious" guidance. "Sean, function X obviously should be in file Y" is perfectly acceptable here. 😄 

If we can, let's leave styling review out for now. Maybe one day this project will get a face lift, but I'm probably not the right person to invest in that - we're focused on functionality ATM.

### Disclaimer
It should be noted - this is not intended to be example code that could be mimicked for a production search application - this is NOT SECURE, and merely demonstrates how changing API keys based on a persona from an identity index can provide different views into a search application's data. In a real setup, the management of API keys would be done by a backend server that sits between the browser and Elasticsearch.

dls-internal-search


example-apps/internal-knowledge-search/app-ui/elastic-search-application-client-8.9.0-2.tgz
seanstory, line 1: This is a stop-gap until https://github.com/elastic/search-application-client/pull/33 is merged and released

example-apps/internal-knowledge-search/app-ui/src/App.tsx
```diff
@@ -21,8 +21,8 @@ function App() {
                         <SidebarMenu/>
                         <main className="flex-1 p-6">
                             <Routes>
-                                <Route path="/" element={<SearchPage/>}/>
-                                <Route path="/settings" element={<SettingsPage/>}/>
+                                <Route path="/" element={<SettingsPage/>}/>
+                                <Route path="/search" element={<SearchPage/>}/>
```
seanstory, line 25: This is a hack. Since I'm generating api keys dynamically now, but on settings page load, I force you to start at the settings page. 

It would be better to have the API Key generation logic be resuable between both the search and settings pages, but it wasn't clear to me what the right pattern/location for the code would be. I'd welcome guidance and/or examples on the right way to structure "library" code in react.

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
```diff
@@ -1,23 +1,222 @@
...
+                method: "POST",
+                headers: {
+                    "Authorization": "Basic " + btoa(appUser + ":" + appPassword),
+                    "Content-Type": "application/json"
+                },
+                body: JSON.stringify({
+                    "size": 30 //TODO: fix magic number. This is just how many identities I have in my data
+                })
+            });
+            const jsonData = await response.json();
```
seanstory, line 29: I'm happy to DRY up this pattern of

```
const path = ...
const response =  fetch(path, ...
const jsonData = await respons.json();
```
if there's a suggestion on where a "client" lib/class should live and how to expose it.

example-apps/internal-knowledge-search/app-ui/.env
```diff
@@ -1,3 +1,4 @@
-SEARCH_APP_NAME=some-search-application
-SEARCH_APP_API_KEY=xxxxxxxxxxxxxxxxxxx
-SEARCH_APP_ENDPOINT=https://some-search-end-point.co
\ No newline at end of file
+REACT_APP_SEARCH_APP_NAME=changeme
+REACT_APP_SEARCH_APP_ENDPOINT=http://localhost:9200
+REACT_APP_SEARCH_USER=elastic
+REACT_APP_SEARCH_PASSWORD=changeme
```
miguelgrinberg, line 4: I'm sorry if I'm misunderstanding something here, but are we asking people to send their account password to the front end? I can understand that it is convenient to have the front end do everything without a back end, but this is not a practice that should be encouraged, I think?

example-apps/internal-knowledge-search/app-ui/.env
```diff
@@ -1,3 +1,4 @@
-SEARCH_APP_NAME=some-search-application
-SEARCH_APP_API_KEY=xxxxxxxxxxxxxxxxxxx
-SEARCH_APP_ENDPOINT=https://some-search-end-point.co
\ No newline at end of file
+REACT_APP_SEARCH_APP_NAME=changeme
```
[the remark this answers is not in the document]
seanstory: These are on purpose. This file has 2 over-arching changes:
1. changing the prefix of the variables to be `REACT_` which is apparantly necessary. Lio has made this same change in an isolated PR: https://github.com/elastic/elasticsearch-labs/pull/142
2. removing the API config to replace it with a username/password combo, since API keys cannot be used to create other API keys. See [this slack thread](https://elastic.slack.com/archives/C0D8ST60Y/p1703017952223129).

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
```diff
@@ -1,23 +1,222 @@
...
+        return jsonData.encoded
+    }
+
+    const [searchPersonaOptions, setSearchPersonaOptions] = useState(["admin"]);
+
+    useEffect(()=>{
+        (async()=>{
+            const fetchedPersonas = await fetchPersonaOptions()
+            setSearchPersonaOptions(fetchedPersonas)
+            if (searchPersonaAPIKey == "missing") {
```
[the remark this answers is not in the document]
seanstory: Ok sure! I just saw this as the "on load hook", and so put everything in here I wanted to happen on load. Didn't even occur to me that I could define another one to keep them separate.

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
```diff
@@ -1,23 +1,222 @@
...
+        setIsOpen(!isOpen);
+    };
+
+    const handlePersonaChange = async (value: string) => {
+        updateSearchPersona(value, await createPersonaAPIKey(value));
+    };
+
     const handleSave = () => {
-        dispatch(updateSettings({appName, apiKey, searchEndpoint}));
+        dispatch(updateSettings({appName, appUser, appPassword, searchEndpoint, searchPersona, searchPersonaAPIKey}));
```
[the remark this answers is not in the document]
seanstory: AFAIK, it's redundant. This predates my work, so @timgrein may have a different reason, but my guess was just wanting to give the user a button to click that confirms that their data is saved, but to not rely on them having to click it?

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
```diff
@@ -1,23 +1,222 @@
...
+        })()
+    },[])
+
+    const [isOpen, setIsOpen] = useState(false);
+
+    const toggleDropdown = () => {
+        setIsOpen(!isOpen);
+    };
+
+    const handlePersonaChange = async (value: string) => {
```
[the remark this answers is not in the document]
seanstory: > Instead of creating an API key at the same time as changing the persona, it'd be cleaner to update the Persona, and then have an effect that reacts to a Persona change to create a new API key.

I don't think I grok the distinction. Is there a way to "watch" persona for a change, other than to call something from it's update function? Like, is there an event that is triggered that I can listen for when a model changes?

> Even better would be to store a separate API key per persona in the state so you only ever create one API key per persona instead of recreating every time a persona changes back and forth.

Fair. I'd considered adding such a thing to the SearchApplicationSettings model, but felt like it was just getting too big. SearchPersonas probably needs to be its own model, and SearchApplicationSettings just references the ID/Name of a single persona at a time.

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
```diff
@@ -51,17 +250,34 @@ const SearchApplicationSettings: React.FC = () => {
 
             <div className="text-left mb-6 p-4 border rounded bg-gray-50">
                 <label
-                    htmlFor="apiKey"
+                    htmlFor="appUser"
                     className="block text-sm font-medium mb-1 text-gray-700"
                 >
-                    API Key:
+                    Application Elasticsearch Username:
```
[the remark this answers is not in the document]
seanstory: Because it's the username for the user of the Search Application. Which is an Elasticsearch thing, so "Elasticsearch username" would technically be correct, but I wanted to make clear that this is the user executing a search against the Search Application (in the event that another search persona is not selected). WDYT?

example-apps/internal-knowledge-search/app-ui/.env
seanstory: Totally agree. See this slack thread: https://elastic.slack.com/archives/C01795T48LQ/p1702066837681769?thread_ts=1702066380.039599&cid=C01795T48LQ

specifically:
>  I don’t think we want to build an e2e app that is going to be the skeleton a customer could start from. But we want to give them some examples, using the sample app that exists already

and

> > An example application that customers could actually run with is not as small a thing.

> This sounds to me like too big effort to start with. 

Basically it came down to scope, and the fact that this isn't intended to be used for production, but to simply demo on fake data. Once this PR is in a clean place, I'll be starting on a blog/tutorial that will reference this example, but discuss how insecure it is and how a real architecture for DLS would need a backend server between Elasticsearch and the browser that could do some of these elements for you securely.


> are we asking people to send their account password to the front end?

Even with the above context, I'd planned to use only API keys. Unfortunately, there's a 3-year-old "known issue" where ES API keys cannot be used to create API keys for other users/privileges.  See this slack thread: https://elastic.slack.com/archives/C0D8ST60Y/p1703017952223129
So I was faced with either switching to basic auth, or introducing something like JWT as an additional realm and part of setup. And again since this isn't indended to be used for production but just as an example/demo, the extra scope does not seem worth it IMO.

miguelgrinberg: The issue I have is that this is not demonstrating something that has any practical value. If you are going to show how to do something in a bad way that nobody should use, then why show it in the first place?

I have no doubt there are reasons why this is this way, but the reality is that the decision of wether this is supposed to be used in production or not is not ours to make. People who download this example will make this decision, and if you give them something that is badly designed but works, the majority of people will use it without any concerns, because they'll assume that if we do it we think it's okay for them to do it too.

seanstory: > The issue I have is that this is not demonstrating something that has any practical value. If you are going to show how to do something in a bad way that nobody should use, then why show it in the first place?

I think there's some distance between "not intended for production" and "no practical value". 🙂 
The issue that we've seen is that DLS is really confusing for our Field folks. Despite a LOT of documentation we've written on the subject, they struggle to understand the moving pieces enough to be able to put together a POC. The goal here is to provide something similar to [this existing demo](https://esre-openai-sample-app.prod-3.eden.elastic.dev/), but populated by dynamic search personas, rather than hardcoded ones. CC @serenachou and @danajuratoni in case y'all would like to add some more context.

>  the majority of people will use it without any concerns

I think that's a bit of a leap, especially if we go out of our way to flag that this is not production-ready or secure. We could definitely go further than I have so far though - would it assuage some of your concerns if we added a banner to the top of the search experience to note that this is not for production? And/or if we added a large disclaimer at the top of the example app's README?

serenachou: For this knowledge base app - since it's not meant for production in current scope, let's clearly document the password strategy here with a warning that customers should look to more secure methologies.

miguelgrinberg: I think what you are implicitly saying is that it is okay to expose secrets on apps that are not production apps. I disagree. My position is that a solution that requires Elastic credentials to be used from a front end has no practical value at all. It does not matter that we think this is not a production-ready solution. There is no reason to share a bad solution when there is a correct and safe solution that we could promote instead.

But in any case, I shared my thoughts and you have my feedback. Your change will make this already bad situation worse, because in the current version only an API key is shared and now you are changing it to share the account password. Exposing an API key is still bad, but not as bad as sharing the password, which provides full access to everything.

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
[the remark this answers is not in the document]
sphilipse: > I don't think I grok the distinction. Is there a way to "watch" persona for a change, other than to call something from it's update function? Like, is there an event that is triggered that I can listen for when a model changes?

Yes! `useEffect`s dependency array allows you do this:

```
useEffect(async () => {
  const fetchApiKey = async () => { 
     return await createPersonaApiKey(value);
  }
  const apiKey = fetchApiKey();
  updateApiKey(apiKey);
}, [searchPersona]);
```

Note that here we do define an async function inside of this effect and then call it, because we care about the result of that async await call inside of this effect.

> Fair. I'd considered adding such a thing to the SearchApplicationSettings model, but felt like it was just getting too big. SearchPersonas probably needs to be its own model, and SearchApplicationSettings just references the ID/Name of a single persona at a time.

Yeah I figured and this is a fine compromise for an app like this.

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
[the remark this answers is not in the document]
sphilipse: I'd either get rid of the save button, or make the state behave the way you'd expect it (ie: all-local and only pushed to global/persistent state on save).

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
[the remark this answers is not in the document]
sphilipse: `Search Application User` might be clearer then? It's not a big deal anyway and I'm not sure my intuitions are correct here, I just stumbled over it because it seems very awkwardly phrased.

example-apps/internal-knowledge-search/app-ui/.env
sphilipse: I'm gonna have to agree with @miguelgrinberg here. I've seen too many folks just run these types of examples as-is even in prod or nearly-prod because "hey that's how the example app works" and people often don't have a good mental model of how frontend and backend are separated.

It's probably worth it adding a very barebones backend to filter these API calls through, and have that backend contain the Elasticsearch config. Note that several other example apps in this repo already do that, including the Search Tutorial and the Chatbot RAG App.

sphilipse: It might be better to do that in a separate PR, to prevent the scope from blowing up in this PR though.

example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx
```diff
@@ -1,23 +1,222 @@
...
+            })
+        })
+        const jsonData = await response.json()
+        return jsonData.encoded
+    }
+
+    const [searchPersonaOptions, setSearchPersonaOptions] = useState(["admin"]);
+
+    useEffect(()=>{
+        (async()=>{
```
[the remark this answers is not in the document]
seanstory: If I do as you suggest, my IDE gives me an error:
```
'await' expressions are only allowed within async functions
```
and if I make it:
```
useEffect( async () => {...})
```
it complains:
```
Type 'Promise<void>' is not assignable to type 'void | Destructor'
```

Which I think is what the original syntax was getting around, right? Is there another way to be able to call async functions (`fetchPersonaOptions()` and `createPersonaAPIKey()`) in `useEffect()`?

example-apps/internal-knowledge-search/app-ui/.env
seanstory: > I think what you are implicitly saying is that it is okay to expose secrets on apps that are not production apps

I think it would be more accurate to say that we think it's ok to expose demo/local secrets on a demo/local system. I think we're all in agreement that it's not ok to risk "real" credentials or customer credentials.

> There is no reason to share a bad solution
> Your change will make this already bad situation worse

Let's try to limit the value-judgement. We're all on the same team :)
I see where you're coming from - let's work to find a compromise between technical best practices and real-world time/resource constraints.

> there is a correct and safe solution that we could promote instead.

There is. @serenachou @danajuratoni , I'd expressed my concerns with this approach early on, and now Miguel and Sander have both also expressed strong reservations. Can we re-evaluate if we can afford to invest the time to build a proper backend? 

> Note that several other example apps in this repo already do that, including the Search Tutorial and the Chatbot RAG App.

That's news to me. I dig into these a little next week to try to assess how much effort it would be to shift this application to one of those. I believe these are also being leveraged in EDEN, so I don't think changing architecture to something else we've proven out would put too much burden on anyone downstream.

> It might be better to do that in a separate PR, to prevent the scope from blowing up in this PR though.

I'd probably close this PR, but keep working in the same branch?  But let's get aligned on direction first before we get into the weeds of implementation. 👍

danajuratoni: The main 2 goals of this initiative are 

1. provide customers and Field with guidance how to set up DLS for POCs. We've seen multiple instances where customers and Field colleagues were not able to successfully set up DLS after many days of research. Providing an example on how this can be achieved, has immense value.
2. additionally, the example app should be an easy to reuse tool that @lio-p can use to create eden demos.

For DLS specifically, this example successfully shows how permissions should be set up. However, the goal wasn't to show a best practice for an entire search application architecture. The security concerns voiced out in this thread are valid concerns, and the question I see raised is: _does the value of this example app that zooms in on setting up DLS justify deferring / separately addressing generic search experience design & development "best practices" concerns?_ cc: @serenachou 
From the DLS POV, my strong opinion is that we need an example of how DLS works e2e. And this was brought up numerous times already. Sean's work is a great starting point we can use in our discussions, and this initiative should be continued to also showcase our recommendation for designing a search experience, as @serenachou and @radhapolisetty prioritize.

@miguelgrinberg a followup question here is - is the goal to have all example apps in elasticsearch-labs follow specific design standards for search experiences? If so, what are these?

miguelgrinberg: @danajuratoni Since this is a public repository where people come to get examples and ideas to help them build their own projects, I think it is reasonable to expect that we will be promoting safe and industry accepted patterns. I'm not setting a high bar here, just that we should follow standard practices, especially with regards to security.

The version of this application as modified in this PR asks the user to configure their Elastic account password in their front end, and uses it from the front end in a way that is intended to only be used in a server-to-server communication. This is unacceptable, we should never promote any applications that require having sensitive information available in the browser. Even if we feel this is just a demo, not for production, etc. the risk is too high, because people are not going to read the disclaimers and all the "buts" that we can come up with and will copy the insecure solution and run with it no matter the risk.

sphilipse: Hi @miguelgrinberg, I'm going to pick this up and add a quick backend to this PR. Checking to make sure we're aligned on what we'd want in this demo: I think adding authentication and SSL to this backend would be overkill, but happy to hear your thoughts.

miguelgrinberg: Hi @sphilipse, yes, I would also consider SSL and authentication overkill for a demo, in fact we don't have them in the search tutorial and RAG examples either.

sphilipse: Okay cool, I added a backend in https://github.com/elastic/elasticsearch-labs/pull/141/commits/3440e55f59b397b63612283b6e2f39ee1501d6ac, let me know if that matches your expectations @miguelgrinberg and then I think this should be close to ready for merging.

example-apps/internal-knowledge-search/api/app.py
```diff
@@ -0,0 +1,185 @@
...
+        search_app_name = request.args.get("app_name")
+        search_app = elasticsearch_client.search_application.get(
+            name=search_app_name)
+        return search_app['indices']
+
+    except Exception as e:
+        current_app.logger.warn(
+            "Encountered error %s while fetching personas, returning default persona", e
+        )
+        return ["admin"]
```
seanstory, line 88: This messaging looks like copy-pasta from above. Should probably just raise a 500 if you can't read the indices list from the search application?

example-apps/internal-knowledge-search/api/app.py
```diff
@@ -0,0 +1,185 @@
...
+        current_app.logger.warn(
+            "Encountered error %s while fetching api key", e)
+        raise e
+
+
+@app.cli.command()
+def create_index():
+    """Create or re-create the Elasticsearch index."""
+    basedir = os.path.abspath(os.path.dirname(__file__))
+    sys.path.append(f"{basedir}/../")
```
seanstory, line 181: What's this for?

example-apps/internal-knowledge-search/app-ui/src/config/documentsToSearchResultMappings.json
```diff
@@ -1,6 +1,6 @@
 {
-  "search-mongo": {
-    "title": "name",
+  "search-spo": {
+    "title": "Title",
```
seanstory, line 3: Probably want to undo this, to leave mongo documented?
The bottom of this file has an object for sharepoint

example-apps/internal-knowledge-search/app-ui/src/store/slices/searchApplicationSettingsSlice.ts
```diff
@@ -5,22 +5,40 @@ import {SearchApplicationSettingsModel} from "../../models/SearchApplicationSett
...
     searchPersona: "admin",
     searchPersonaAPIKey: "missing"
 };
 
-const searchApplicationSettingsSlice = createSlice({
+export const searchApplicationSettingsSlice = createSlice({
     name: 'searchApplicationSettings',
     initialState,
     reducers: {
+        updateAppName: (state, action: PayloadAction<string>) => {
```
seanstory, line 18: Ah, thanks for doing this. You'd left a comment for me to address this and I'd forgotten to.

example-apps/internal-knowledge-search/app-ui/src/store/slices/searchApplicationSettingsSlice.ts
```diff
@@ -5,22 +5,40 @@ import {SearchApplicationSettingsModel} from "../../models/SearchApplicationSett
...
+            state.appName = action.payload;
+            return state;
+        },
+        updateIndices: (state, action: PayloadAction<string[]>) => {
+            state.indices = action.payload;
+            return state;
+        },
         updateSettings: (state, action: PayloadAction<SearchApplicationSettingsModel>) => {
             return action.payload;
         },
```
seanstory, line 28: Is this needed anymore, if we can update things individually?

example-apps/internal-knowledge-search/env.example
```diff
@@ -0,0 +1,16 @@
+# Make a copy of this file with the name .env and assign values to variables
+
+# Your Elaticsearch credentials
+ELASTICSEARCH_URL=http://localhost:9200
+ELASTIC_API_KEY=
+
+# If you prefer using a cloud ID
+# ELASTIC_CLOUD_ID=
+
+# If you prefer using username and password
```
seanstory, line 10: Don't you need to use username/password? You can't create API keys if you authenticated with an API key.
Or maybe we  should just have in this comment: 
```
# If you're using Document Level Security, or if you just prefer using username and password
```

example-apps/internal-knowledge-search/env.example
```diff
@@ -0,0 +1,16 @@
...
+# ELASTIC_USERNAME=
+# ELASTIC_PASSWORD=
+
+SEARCH_APP_NAME=changeme
+SEARCH_APP_USER=elastic
+SEARCH_APP_PASSWORD=changeme
```
seanstory, line 16: Are these redundant?

example-apps/internal-knowledge-search/requirements.txt
```diff
@@ -0,0 +1,52 @@
...
+blinker==1.6.2
+    # via flask
+build==1.0.3
+    # via pip-tools
+certifi==2023.7.22
+    # via elastic-transport
+click==8.1.7
+    # via
+    #   flask
+    #   pip-tools
```
seanstory, line 16: some of these deps might not be needed here. Not the end of the world if this is just copy-pasted from another example app, but I'd think we wouldn't need click, jinja, markup, etc.

seanstory (commented):
Few little things, but overall I think this is great. Thanks so much!

example-apps/internal-knowledge-search/requirements.txt
sphilipse: These are all dependencies of dependencies. Requirements.txt was auto-generated from requirements.in, which I culled.

example-apps/internal-knowledge-search/env.example
sphilipse: Yup, removed

example-apps/internal-knowledge-search/env.example
sphilipse: Done

example-apps/internal-knowledge-search/api/app.py
sphilipse: Copy-paste leftover

example-apps/internal-knowledge-search/api/app.py
sphilipse: Done

PR #142, opened 2023-12-21
Rename env variables to be used in React

lio-p:


example-apps/internal-knowledge-search/app-ui/.env
```diff
@@ -1,3 +1,3 @@
-SEARCH_APP_NAME=some-search-application
-SEARCH_APP_API_KEY=xxxxxxxxxxxxxxxxxxx
-SEARCH_APP_ENDPOINT=https://some-search-end-point.co
\ No newline at end of file
+REACT_APP_SEARCH_APP_NAME=some-search-application
+REACT_APP_SEARCH_APP_API_KEY=xxxxxxxxxxxxxxxxxxx
```
miguelgrinberg, line 2: Correct me if I'm wrong, but this would make the API key available in the browser, which is a really bad idea security-wise, and not something we should promote or recommend.

lio-p: Yes you're correct, but actually the current app already does expose it. But this is the method documented in the [Elasticsearch documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-application-security.html).

This PR just fixes a bug that prevent developers to pass those parameters using an `.env` file. If not, those parameters can be set through the [web app itself](https://github.com/elastic/elasticsearch-labs/blob/da31f8269322ab44dfca890849970bae7f1148e5/example-apps/internal-knowledge-search/app-ui/src/components/SearchApplicationSettings.tsx#L19). 

Maybe we should add a disclaimer in the README about the security risks and point to the [Elasticsearch documentation]((https://www.elastic.co/guide/en/elasticsearch/reference/current/search-application-security.html) for more details?

miguelgrinberg: Interesting that we recommend this method, as it is very risky and depends on people having an understanding of what they're doing when generating their keys, which isn't always the case.

Even with a scoped API key, there is risk of a bad actor using the key to generate unexpected expenses for the application owner, or render the search functionality unusable by making enough calls to hit the rate limits, so this is still a bad idea.

At the very least there should be a very big warning in the readme explaning the risk.

lio-p: I added a disclaimer in the Readme file,  please let me know what you think.

PR #173, opened 2024-01-25
Tokenization notebook

maxjakob:
Add notebook to show users how tokenization works in the context of semantic search and explain why and how users need to chunk text. (Follow up on a [forum thread](https://discuss.elastic.co/t/embedding-token-size-limit-for-elser2-model/351403/11)).

### TODO
- [x] Refine recommendation at the end
- [x] Remove output of some cells
- [x] Read text file from the web
- ~Add reference to overlap documentation~
- ~Comment on the fact that it can happen that chunks start with a `##foo` token.~


notebooks/search/lorem-ipsum.txt
joshdevins, line 1: Can we use some existing text already in the repo? We have a few datasets. Or add this to the root datasets directory maybe.

notebooks/search/tokenization.ipynb
```diff
@@ -0,0 +1,330 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
```
joshdevins, line 63: Clear cell output.

joshdevins (changes requested):
It would be good to add a stride/overlap when we chunk since this is a best practice. With ELSER, we recommend 50% overlap/256 token stride.

notebooks/search/tokenization.ipynb
```diff
@@ -0,0 +1,322 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Handling long texts\n",
+    "\n",
+    "We saw how to count the number of tokens using the tokenizers from different models. ELSER uses the BERT tokenizer, so when using `.elser_model_2` it internally splits the text with this method.\n",
+    "\n",
+    "Currently there is a limitation that [only the first 512 tokens are considered](https://www.elastic.co/guide/en/machine-learning/8.12/ml-nlp-limitations.html#ml-nlp-elser-v1-limit-512). To work around this, we can first split the input text into chunks of 512 tokens and feed the chunks to Elasticsearch separately. Actually, we need to use a limit of 510 to leave space for the two special tokens (`[CLS]` and `[SEP]`) that we saw.\n",
+    "\n",
+    "Furthermore, it is best practice to make the chunks overlap (**TODO add reference**). With ELSER, we recommend 50% token overlap (i.e. a 256 token stride)."
```
maxjakob, line 210: @joshdevins Do you have a reference that we can link here?

miguelgrinberg:
How do you feel about making this notebook run cleanly under `nbtest`?

maxjakob:
> How do you feel about making this notebook run cleanly under nbtest?

Definitely. Runs fine for me locally:
```sh
$ nbtest notebooks/search/tokenization.ipynb
Running notebooks/search/tokenization.ipynb... OK
```

@miguelgrinberg can you confirm that nbtest always expects me to set env variables for Elastic Cloud access even if my notebook does not require it?

miguelgrinberg:
There shouldn't be a need to set env vars if they are not used by the notebook. Are you getting any error(s) that I can see?

Also, great that the notebook runs cleanly!

notebooks/search/tokenization.ipynb
```diff
@@ -0,0 +1,322 @@
...
+     },
+     "execution_count": 10,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "tokens = bert_tokenizer.encode(long_text)[1:-1]  # exclude special tokens at the beginning and end\n",
+    "chunked = [\n",
+    "    bert_tokenizer.decode(tokens_chunk)\n",
```
davidkyle, line 278: Encoding the text involves a step that normalises the text to Unicode NFD form and then strips non spacing marks (Unicode category Mn).  

Decoding does a reverse look up on the BERT vocabulary file and joins the compound element (those starting with `##`). 

Decoding isn't perfect as those non spacing marks are lost and so may appear different to the original text.

This might be a level of detail too far for most readers but it's worth mentioning. Maybe something like:

"Tokenisation involves a normalisation step and strips non spacing marks. If decoding is implemented as a reverse lookup from token Ids to vocabulary entries those stripped marks will not be recovered resulting in decoded text that could slightly different to the original."

Again I'm not sure we need this level of detail

davidkyle (approved):
LGTM

This is brilliant we should have done it ages ago.

notebooks/search/tokenization.ipynb
joshdevins: I do not. Maybe @qherreros does? I don't think it's in public docs anywhere, just in our internal benchmarking.

joshdevins (approved):

notebooks/search/tokenization.ipynb
qherreros: The best reference we have on this subject is in [internal documentation](https://docs.google.com/spreadsheets/d/1ojhOXu4rpl8dxKVqGUMfkYmt6vk5006CI7duFVimV5k/edit#gid=0).
I found this paper which is explicitly measuring improvements with overlaps, but it's in [encoder/decoder architecture](https://direct.mit.edu/tacl/article/doi/10.1162/tacl_a_00547/115346/Efficient-Long-Text-Understanding-with-Short-Text), fusing chunks in the decoder. I still think their conclusion can be applied to more than just FiD architecture.

notebooks/search/tokenization.ipynb
maxjakob: Interesting point. This seems to only be a problem for BERT though. The E5 tokenizer preserves the nonspacing marks:
```py
latvian_name = "Da\u0305vis"
print(latvian_name)
print(bert_tokenizer.decode(bert_tokenizer.encode(latvian_name)))
print(e5_tokenizer.decode(e5_tokenizer.encode(latvian_name)))
```
```
Da̅vis
[CLS] davis [SEP]
<s> Da̅vis</s>
```
If we recommend using ELSER=BERT for English-only and E5 for multi-lingual, I lean towards omitting this point here.

notebooks/search/tokenization.ipynb
```diff
@@ -0,0 +1,322 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "The lives of two mob hitmen, a boxer, a gangster and his wife, and a pair of diner bandits intertwine in four tales of violence and redemption.\n",
```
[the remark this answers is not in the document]
maxjakob: A Royale with cheese 🍔

notebooks/search/tokenization.ipynb
maxjakob: Alright, I went without a reference and this generic statement: _in practice we often see improved performance when using overlapping chunks_

maxjakob:
> What you need to do now is add a Makefile, and then reference the Makefile in the top-level Makefile, so that this gets included when I run make test from the top!

@miguelgrinberg Added to the existing Makefile: 4f6dfaab5a8f457c544b758ea7599ff0a829186f

PR #191, opened 2024-02-22
Added modal status check, delete index before creating index

saarikabhasi:
Found as part of [8.13 Search labs QA ticket](https://github.com/elastic/enterprise-search-team/issues/6957) 

Tests were failing, as it takes a while for modal to be started with Eland tool. Added modal status check similar to ELSER v2 modal check. 
Also add delete index step before creating index.


miguelgrinberg:
Waiting the correct amount of time has proven very difficult. I never had issues with this notebook (or actually any of the others uploading models via eland), but the change seems reasonable, so let's go with it.

PR #196, opened 2024-03-04
Update LangChain imports to use `langchain-elasticsearch`

maxjakob:
the new partner integration package:
https://pypi.org/project/langchain-elasticsearch/

In many cases the `elasticsearch` package does not need to be explicitly imported any more because `langchain-elasticsearch` depends on it and it's not directly imported.

I admit to running only 2 notebooks to verify that things work, but I triple-checked all spellings ;)


joemcelroy:
also could you update https://github.com/elastic/elasticsearch-labs/tree/main/example-apps/chatbot-rag-app too please?

PR #200, opened 2024-03-08
Updated langchain-core dependency to resolve conflict.

hemantmalik:
I updated langchain-core version in requirements.txt to resolve dependency conflict that was failing the docker build. Also updated the comment in the env.example to highlight the code is tested for Mistral AI Large deployed through Microsoft Azure.

### DIFF
--- a/example-apps/chatbot-rag-app/env.example
+++ b/example-apps/chatbot-rag-app/env.example
@@ -35,7 +35,7 @@ ES_INDEX_CHAT_HISTORY=workplace-app-docs-chat-history
 # VERTEX_REGION=
 # GOOGLE_APPLICATION_CREDENTIALS=
 
-# Uncomment and complete if you want to use Mistral AI
+# Uncomment and complete if you want to use Mistral AI Large depployed through Microsoft Azure
 # LLM_TYPE=mistral
 # MISTRAL_API_KEY=
 # MISTRAL_API_ENDPOINT=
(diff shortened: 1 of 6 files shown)


example-apps/chatbot-rag-app/env.example
```diff
@@ -35,7 +35,7 @@ ES_INDEX_CHAT_HISTORY=workplace-app-docs-chat-history
 # VERTEX_REGION=
 # GOOGLE_APPLICATION_CREDENTIALS=
 
-# Uncomment and complete if you want to use Mistral AI
+# Uncomment and complete if you want to use Mistral AI Large depployed through Microsoft Azure
```
miguelgrinberg, line 38: I don't understand this. Nothing in this change has a constraint on a specific endpoint or model, so why mention that here?

hemantmalik: That's a good point. I have yet to test the code against a direct Mistral AI endpoint. The mention was there to clarify that this was tested with the Mistral AI Large deployed on Azure in case there is anything different about direct Mistral AI endpoints. We can change it back.

miguelgrinberg: If we want to make a disclaimer that this was tested only on a specific subset, I'd suggest doing so in the readme file, where the documentation is, maybe even include deployment instructions and example values for the endpoint and model environment variables.

The code change is really not dependent on a specific deployment method or model though, so I suggest being more specific, like for example adding a note along the lines of "While all Mistral AI models should be supported, we have only tested the Mistral AI Large model deployed on Azure".

hemantmalik: Thanks, Miguel. I will update the Readme

miguelgrinberg (approved):

PR #201, opened 2024-03-11
Add esql getting started notebook

leemthompo:
<a target="_blank" href="https://colab.research.google.com/github/elastic/elasticsearch-labs/blob/38362a9ed1583afe0a5197214c7fc144125e1b66/notebooks/esql/esql-getting-started.ipynb">Open In Colab</a>

Ports getting started to Python.

### TODO

- [x] Write introduction

### DIFF
--- a/notebooks/esql/esql-getting-started.ipynb
+++ b/notebooks/esql/esql-getting-started.ipynb
@@ -6,7 +6,53 @@
     "id": "yfVx1CGq7dZU"
    },
    "source": [
-    "<a target=\"_blank\" href=\"https://colab.research.google.com/github/leemthompo/notebook-tests/blob/main/esql-812-tutorial.ipynb\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+    "<a target=\"_blank\" href=\"https://colab.research.google.com/github/elastic/notebooks/esql/esql-getting-started.ipynb.ipynb\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Getting started with ES|QL\n",
+    "\n",
+    "In this notebook you'll learn the basics of the Elasticsearch Query Language (ES|QL).\n",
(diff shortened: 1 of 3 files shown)


miguelgrinberg:
@leemthompo are there any differences between cloud and docker image with regards to this? The tests run against a self-hosted minimal container, so the behavior might  be different than running against cloud.

Another issue you are going to hit is that the notebook is also going to run against 8.11 and 8.12. Sounds like I'll need to figure out a mechanism to specify which versions a notebook needs to run on.

miguelgrinberg:
The easy option is to add this notebook to the exempt list in [this file](https://github.com/elastic/elasticsearch-labs/blob/main/bin/find-notebooks-to-test.sh). The hard option would be to figure out this failure and adapt the notebook to work under Docker as well. And I need to expand the file I linked above to be version aware and provide different lists of notebooks to skip for each given stack release, which I'm planning to anyway.

leemthompo:
> The hard option would be to figure out this failure and adapt the notebook to work under Docker as well. 

I'm not sure if multi-node Docker clusters solves the issue, but unfortunately I ran out of time to confirm this today.

miguelgrinberg:
I don't think we can stand up a multi-node Docker instance within the resource constrained GitHub Actions. I think the best course of action for now is to skip this notebook when testing.

notebooks/esql/esql-getting-started.ipynb
```diff
@@ -0,0 +1,1402 @@
...
+    "Once logged in to your Elastic Cloud account, go to the [Create deployment](https://cloud.elastic.co/deployments/create) page and select **Create deployment**. Leave all settings with their default values."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Install packages and import modules\n",
+    "\n",
+    "To get started, we'll need to connect to our Elastic deployment using the Python client. Because we're using an Elastic Cloud deployment, we'll use the **Cloud ID** to identify our deployment.\n",
```
szabosteve, line 55: It might be helpful to add a note about getpass here.

szabosteve (commented):
Awesome notebook! It has a great flow and is easy to follow. I tested it end-to-end on a 8.14 cluster (es-py 8.12.0) and it worked flawlessly.

miguelgrinberg (approved):

PR #203, opened 2024-03-12
LTR notebook: fix conclusion

maxjakob:
Small content fix (important since now that we point our [docs](https://www.elastic.co/guide/en/elasticsearch/reference/8.13/learning-to-rank.html#learning-to-rank-models) at this notebook).


miguelgrinberg (approved):
Looks good, but I'm curious to know where did "The Phantom Menace" go that does not show at all in the list.

joemcelroy (approved):

PR #204, opened 2024-03-12
Adds Cohere inference service notebook

szabosteve:
## Overview

Related to https://github.com/elastic/search-docs-team/issues/71
This PR adds a Jupyter notebook that contains an end-to-end example of using the Inference API with the Cohere service.
It also makes some changes in the OpenAI notebook to make it more exact.

### Preview

[Open In Colab](https://colab.research.google.com/github/elastic/elasticsearch-labs/blob/szabosteve/inference-cohere-notebook/notebooks/integrations/cohere/inference-cohere.ipynb)


miguelgrinberg:
@szabosteve Hey, you may want to add this notebook to the testing exception list on [this file](https://github.com/elastic/elasticsearch-labs/blob/main/bin/find-notebooks-to-test.sh), since we are unable to run tests for notebooks that depend on third-party services at this time.

szabosteve:
Thanks for the info, @miguelgrinberg! I've added the notebook to the exempt list via [d760dd4](https://github.com/elastic/elasticsearch-labs/pull/204/commits/d760dd4949de1e7f78dbc79cea3a93033cba6272).

szabosteve:
@joemcelroy Yup, I did it via [6ece101](https://github.com/elastic/elasticsearch-labs/pull/204/commits/6ece101a523e9cefdd922c2b8a717d0e1b714a57). It seems the pre-commit hook is also passed.

leemthompo:
Looks great, nice and clean! 👏 

One open question about needing a **paid Cohere account**. If there is a workaround to use trial API keys, it might be worth considering. This would maximize accessibility. We already set the bar pretty high because you need an Elastic Cloud account.

szabosteve:
Thanks for the review! I have a workaround that is a bit more complicated but makes the bulk request unnecessary, so users without a paid Cohere account can use it. I proposed adding it to the OpenAI notebook but the decision was to remove that option for more clarity.

joemcelroy (approved):

PR #214, opened 2024-03-25
Changes model_id path parameter to inference_id in openAI and Cohere notebooks

szabosteve:
## Overview

This PR renames the `model_id` path parameter to `inference_id` in the OpenAI and the Cohere notebooks.


miguelgrinberg:
I do not have the history on why this argument was renamed, but I assume people working with 8.12 will see errors if they use the new name? In that case, should we note that `model_id` should be used with old versions?

miguelgrinberg:
@szabosteve there is a `model_id` in [line 247](https://github.com/elastic/elasticsearch-labs/pull/214/files#diff-16f51b238c6225a531a393c86f5087e2f413b51f29a8321d3e6fe1e85080ccd8R247) of the cohere notebook. Does that also need to be changed to `inference_id`?

szabosteve:
@miguelgrinberg No, it doesn't need to be changed. Unfortunately, the inference pipeline processor still uses `model_id`. There is a plan to change it in the 8.14 timeframe.

miguelgrinberg (approved):

PR #217, opened 2024-04-01
Add App Search Engine Exporter Notebook

markjhoy:
This adds an initial notebook that shows how to export an App Search engine to Elasticsearch.

Note for reviewers:

This is an initial version, it is not meant to be a perfect, final version of the App Search engine exporter.
This notebook describes the workflow on how to reindex the engine documents in a new Elasticsearch index.
It makes a conservative assumption that the new index still needs to preserve part of the App Search mappings, with only a few changes:

it adds a synonyms-filter to some of the analyzers to use a the Elasticsearch synonym set exported from App Search
it adds sparse_vector fields to showcase how to do semantic search with ELSER
At this moment, we are not necessarily looking for feedback on whether we should suggest a different mapping.

(based on Ioana's [initial PR](https://github.com/elastic/elasticsearch-labs/pull/212) with various cleanups and additional instructions and rewordings).


notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,884 @@
...
+      "metadata": {
+        "id": "lyc-DcjnvTgH"
+      },
+      "source": [
+        "## Export App Search curations in Elasticsearch\n",
+        "\n",
+        "Next, we will export any curations that may be in our App Search engine.\n",
+        "\n",
+        "To export App Search curations we will use Elasticsearch [query rules](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-using-query-rules.html).\n",
+        "At the moment of writing this notebook Elasticsearch query rules only allow for pinning results unlike App Search curations that also allow excluding results.\n",
```
kderusso, line 161: There are some size limitations with query rules as well, we may want to think about what happens if they have > 100 or > 1000 curations. 100 is a soft limit that can be configured and 1000 is currently a hard limit (we want to look into this more as part of going out of tech preview)

joemcelroy:
i haven't run it but it looks good so far! 

One thing to keep in mind (because im the same spot) is that semantic_text is coming and with that, recommending to use that over setting up a pipeline would probably be a better experience. 

Also an example for autocomplete and suggestions would be good.

markjhoy:
> i haven't run it but it looks good so far!
> 
> One thing to keep in mind (because im the same spot) is that semantic_text is coming and with that, recommending to use that over setting up a pipeline would probably be a better experience.
> 
> Also an example for autocomplete and suggestions would be good.

I definitely like these ideas, and think that a "Part 2" would be a great place for this (as this initial notebook isn't intended to be a complete feature guide, and it's getting long enough now)... unless there's pushback to add these here (specifically autocomplete and suggestions), I'd like to keep the scope as-is for now. LMK

joemcelroy:
one thing to note, CI is failing because its trying to run regression checks on notebooks (running the notebook in CI and validating the output in the notebook matches the cell output committed). To omit from this, add the notebook here https://github.com/elastic/elasticsearch-labs/blob/main/bin/find-notebooks-to-test.sh#L3-L23

markjhoy:
> one thing to note, CI is failing because its trying to run regression checks...

Ah! Thank you! I was going to ask about that in a bit. I'll fix this up.

joemcelroy:
FYI our CI is failing for 8.12 because of the latest eland dep requires 8.13. we are doing a short term fix here for now https://github.com/elastic/elasticsearch-labs/pull/218

ioanatia (approved):
thank you for taking this to the finish line - I reviewed your changes and they make sense to me.

In terms of whether we should use `semantic_text` or not, I think we should do that later once the feature is well released, tested and documented.

miguelgrinberg:
You can now rebase the PR to address the build failure on the 8.12 stack.

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+      "metadata": {
+        "id": "zFIDykq6tpY8"
+      },
+      "source": [
+        "# App Search Engine exporter to Elasticsearch\n",
+        "\n",
+        "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/markjhoy/elasticsearch-labs/blob/markjhoy/add_app_search_export_notebook/notebooks/enterprise-search/app-search-engine-exporter.ipynb)\n",
+        "\n",
+        "<details>\n",
+        "<summary>TODO: before merging, use the \"final\" Colab URL here</summary>\n",
```
kderusso, line 14: 👍

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+        "We can retrieve the field types of an App Search engine using the [Schema API](https://www.elastic.co/guide/en/app-search/current/schema.html) and then construct our mapping."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "# define SOURCE_INDEX and DEST_INDEX which we will continue to reuse; feel free to adjust DEST_INDEX\n",
```
kderusso, line 284: Would it make sense to get the value of `DEST_INDEX` as an input parameter?

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+        "  },\n",
+        "  settings=settings\n",
+        ")"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "# Add `sparse_vector` fields for semantic search\n",
```
kderusso, line 506: We should probably call this out as an optional in the section header, additive step, and link to some documentation on semantic search? Could also be a good call out to link to search labs here. 

It also requires the prereq of an ML node

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+        "  index=DEST_INDEX,\n",
+        "  properties=sparse_vector_fields\n",
+        ")"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "# Setup an ingest pipeline using ELSER\n",
```
kderusso, line 537: We could do this via API commands too, but I think this is fine

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+        "results = elasticsearch.search(\n",
+        "  index=SOURCE_INDEX,\n",
+        "  query=app_search_query_payload[\"query\"],\n",
+        "  highlight=app_search_query_payload[\"highlight\"],\n",
+        "  source=app_search_query_payload[\"_source\"],\n",
+        "  sort=app_search_query_payload[\"sort\"],\n",
+        "  size=app_search_query_payload[\"size\"]\n",
+        ")\n",
+        "\n",
+        "print(results)"
```
kderusso, line 758: Should we pretty print this output?

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+        "\n",
+        "print(f\"Elasticsearch query:\\n{json.dumps(semantic_query, indent=2)}\")\n",
+        "\n",
+        "results = elasticsearch.search(\n",
+        "  index=DEST_INDEX,\n",
+        "  query=semantic_query,\n",
+        "  min_score=20\n",
+        ")\n",
+        "\n",
+        "print(results)\n"
```
kderusso, line 806: Should we pretty-print the results here?

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -0,0 +1,894 @@
...
+        "  index=SOURCE_INDEX,\n",
+        "  query=payload[\"query\"],\n",
+        "  highlight=payload[\"highlight\"],\n",
+        "  source=payload[\"_source\"],\n",
+        "  sort=payload[\"sort\"],\n",
+        "  size=payload[\"size\"],\n",
+        "  min_score=1\n",
+        ")\n",
+        "\n",
+        "print(results)"
```
kderusso, line 867: Shoudld we pretty-print the results here?

kderusso (approved):
Nice work! I left some very minor comments, feel free to take them or disregard.

joemcelroy:
here some steps to run linting locally https://github.com/elastic/elasticsearch-labs/blob/main/CONTRIBUTING.md#pre-commit-hook

notebooks/enterprise-search/app-search-engine-exporter.ipynb
markjhoy: Meh - I'm on the fence about that... I think it's enough to have it as a variable, but, in the markdown above that definition, I'll call it out a bit better.

notebooks/enterprise-search/app-search-engine-exporter.ipynb
markjhoy: I'll keep this here for now to keep things simple, but will point the user in the direction if they want to read / use the API instead.

markjhoy:
👍  I think this is ready to go then 🎉

joemcelroy (approved):

joemcelroy:
one thing to note with using ELSER is make the user aware of the model's token limit. https://www.elastic.co/guide/en/machine-learning/current/ml-nlp-limitations.html#ml-nlp-elser-v1-limit-512

In future this will be less of a problem with semantic_text and autochunking.

PR #222, opened 2024-04-08
Notebook telemetry

miguelgrinberg:


telemetry/telemetry.py
```diff
@@ -0,0 +1,57 @@
...
+    """Enable telemetry for the given elasticsearch client instance."""
+    if "nbtest" in os.environ.get("_", ""):
+        # no telemetry for tests
+        return
+
+    if notebook_name is None:
+        notebook_name = get_notebook_name()
+
+    client.options(
+        headers={**client._headers, "user_agent": f"searchlabs/{notebook_name}"}
```
maxjakob, line 55:
```suggestion
        headers={**client._headers, "User-Agent": f"searchlabs/{notebook_name}"}
```
s/_/-/

notebooks/search/00-quick-start.ipynb
```diff
@@ -135,19 +135,51 @@
...
+   "execution_count": null,
+   "id": "3b04f442-729d-406d-b826-654483498df6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#!curl -O -s https://raw.githubusercontent.com/elastic/elasticsearch-labs/main/telemetry/telemetry.py\n",
+    "!curl -O -s https://raw.githubusercontent.com/miguelgrinberg/elasticsearch-labs/notebook-telemetry/telemetry/telemetry.py\n",
+    "from telemetry import enable_telemetry\n",
+    "\n",
+    "enable_telemetry(client)"
```
maxjakob, line 159: I'm wondering if we should just explicitly add the names in the code. It makes it more obvious what exactly we track, which users might appreciate. And we would remove some of the uncertainty of the extraction failing with future versions.

telemetry/telemetry.py
```diff
@@ -0,0 +1,57 @@
...
+
+import os
+
+
+def get_notebook_name():
+    """Return the name of the running notebook."""
+    name = "unknown"
+
+    # first we try to get the name from the Jupyter environment
+    if os.environ.get("JPY_SESSION_NAME"):
```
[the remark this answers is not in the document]
maxjakob: This one we can test. For Colab and VSCode it would be hard to make ourselves an environment.

maxjakob (approved):
I'm generally fine with this way, but I lean towards adding the notebooks names explicitly.

notebooks/search/00-quick-start.ipynb
miguelgrinberg: My only problem is that it requires more maintenance. Every notebook will have to be hand-edited. The goal of auto-determining the name is that you have a snippet that you copy/paste into a new notebook and you are done. Also a visible name may lead to some people editing the name, so we are going to get some amount of bogus names, probably.

joemcelroy: with the combination of the way we derive the name (which I think will need careful maintenance aswell) and every notebook needs to be edited anyway to include this script, i personally would rather this be explicit.

telemetry/telemetry.py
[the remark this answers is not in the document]
miguelgrinberg: If our telemetry starts showing a large number of "unknown" that would be a good indication that something broke.

Aside from that, as Max suggested, we could build a unit test for Jupyter, but doing it for vscode or Colab would require a complex set up in which we can run a browser.

notebooks/search/00-quick-start.ipynb
miguelgrinberg: I'm not convinced, but that's fine, we can start with explicit names and reconsider in the future if we get annoyed having to carefully update all notebooks without messing up the names.

telemetry/telemetry.py
```diff
@@ -0,0 +1,57 @@
...
+    """Enable telemetry for the given elasticsearch client instance."""
+    if "nbtest" in os.environ.get("_", ""):
+        # no telemetry for tests
+        return
+
+    if notebook_name is None:
+        notebook_name = get_notebook_name()
+
+    client.options(
+        headers={**client._headers, "User-Agent": f"searchlabs/{notebook_name}"}
```
maxjakob, line 55: Just a thought: I'm not sure if there is lots of value in this, but should we do best-effort extraction of the platform (with the code above) and add it to the agent, e.g.
`searchlabs/{notebook_name} jupyter/7.1.2`
`searchlabs/{notebook_name} vscode`
`searchlabs/{notebook_name} colab`

We would learn something about the environments of our users. In my current understanding, the KPI dashboards would keep working the same, but we would gain the possibility to ad-hoc check this (or create another dashboard one day).

miguelgrinberg: I think the correct thing to do here would be to do something like:

    User-Agent: searchlabs/00-quick-start (Jupyter 1.0.0)

Right? This is what browsers do, they stick a lot of platform specific stuff in the comments section of the agent string. I like this idea. I don't think we'll be able to add this extra info on charts though, so I'll ask some questions to see if there is any alternative format that would allow us to include the platform in our charts.

joemcelroy: one thing is to check with clients team on how they will treat the user-agent. I believe "search labs" will be product name and "00-quick-start" will be version name. 

I would like to aggregate on the version name in this example, to see the list of most popular notebooks that have been run. I wouldn't want the aggregation to be split when we include the platform.

joemcelroy (approved):

telemetry/telemetry.py
miguelgrinberg: Yes. I have sent some test requests with the platform. Need to wait about 3 hours for them to show up in our telemetry and see how things look. I'll decide to include it or not depending on that.

miguelgrinberg:
Force-merging to bypass the error caused by the telemetry not being in the main branch yet.

PR #235, opened 2024-05-09
Adds Cohere & Elasticsearch tutorial notebook

szabosteve:
## Overview

Related to https://github.com/elastic/search-docs-team/issues/103.

This PR adds a Jupyter notebook that contains an end-to-end example of all the operations that currently Cohere and Elasticsearch are capable of together.

### Preview

[Open In Colab](https://colab.research.google.com/github/elastic/elasticsearch-labs/blob/szabosteve/cohere-es-notebook/notebooks/integrations/cohere/cohere-elasticsearch.ipynb)

### DIFF
--- a/notebooks/integrations/cohere/cohere-elasticsearch.ipynb
+++ b/notebooks/integrations/cohere/cohere-elasticsearch.ipynb
@@ -9,9 +9,9 @@
     "\n",
     "[![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/elastic/elasticsearch-labs/blob/main/notebooks/integrations/cohere/cohere-elasticsearch.ipynb)\n",
     "\n",
-    "The instructions in this tutorial shows you how to compute embeddings with\n",
+    "This tutorial shows you how to compute embeddings with\n",
     "Cohere using the inference API and store them for efficient vector or hybrid\n",
-    "search in Elasticsearch. This tutorial will use the Python Elasticsearch client\n",
+    "search in Elasticsearch. This tutorial uses the Python Elasticsearch client\n",
     "to perform the operations.\n",
     "\n",
     "You'll learn how to:\n",
@@ -42,7 +42,7 @@
     "- A paid [Cohere account](https://cohere.com/) is required to use the Inference API with \n",
     "the Cohere service as the Cohere free trial API usage is limited.\n",
     "\n",
-    "- Python 3.7 or higher."
+    "- Python 3.7 or later."
    ]
   },
   {
@@ -118,7 +118,7 @@
     "ELASTIC_CLOUD_ID = getpass(\"Elastic Cloud ID: \")\n",
     "\n",
     "# https://www.elastic.co/search-labs/tutorials/install-elasticsearch/elastic-cloud#creating-an-api-key\n",
-    "ELASTIC_API_KEY = getpass(\"Elastic Api key: \")\n",
+    "ELASTIC_API_KEY = getpass(\"Elastic API key: \")\n",
     "\n",
     "# Create the client instance\n",
     "client = Elasticsearch(\n",
@@ -139,8 +139,8 @@
    "source": [
     "## Create the inference endpoint\n",
     "\n",
-    "Create the inference endpoint first. In this example, the inference endpoint will \n",
-    "use Cohere's `embed-english-v3.0` model and the `embedding_type` is set to\n",
+    "Create the inference endpoint first. In this example, the inference endpoint \n",
+    "uses Cohere's `embed-english-v3.0` model and the `embedding_type` is set to\n",
     "`byte`."
    ]
   },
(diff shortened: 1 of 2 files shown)


notebooks/integrations/cohere/cohere-elasticsearch.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "\n",
+    "For this example, you will need:\n",
+    "\n",
+    "- An Elastic deployment with minimum **4GB machine learning node**\n",
+    "   - We'll be using [Elastic Cloud](https://www.elastic.co/guide/en/cloud/current/ec-getting-started.html) for this example (available with a [free trial](https://cloud.elastic.co/registration?utm_source=github&utm_content=elasticsearch-labs-notebook))\n",
+    "   \n",
+    "- A paid [Cohere account](https://cohere.com/) is required to use the Inference API with \n",
+    "the Cohere service as the Cohere free trial API usage is limited.\n",
+    "\n",
+    "- Python 3.7 or higher."
```
marciw, line 45:
```suggestion
    "- Python 3.7 or later."
```

per https://developers.google.com/style/word-list#higher and many others :)

marciw (approved):
LGTM with minor suggestions. 

(Couldn't test w/o Cohere account -- let me know if I should get that set up so I can work through this properly!)

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
```diff
@@ -0,0 +1,562 @@
...
+   "source": [
+    "client.indices.create(\n",
+    "    index=\"cohere-embeddings\",\n",
+    "    settings={\"index\": {\"default_pipeline\": \"cohere_embeddings\"}},\n",
+    "    mappings={\n",
+    "        \"properties\": {\n",
+    "            \"text_embedding\": {\n",
+    "                \"type\": \"dense_vector\",\n",
+    "                \"dims\": 1024,\n",
+    "                \"element_type\": \"byte\",\n",
```
miguelgrinberg, line 213: The `element_type` key is not in the spec. Is this something new that needs to be added?

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
```diff
@@ -0,0 +1,562 @@
...
+    "            \"text_embedding\": {\n",
+    "                \"model_id\": \"cohere_embeddings\",\n",
+    "                \"model_text\": query,\n",
+    "            }\n",
+    "        },\n",
+    "        \"k\": 10,\n",
+    "        \"num_candidates\": 50,\n",
+    "    },\n",
+    "    query={\"multi_match\": {\"query\": query, \"fields\": [\"text\", \"title\"]}},\n",
+    ")\n",
```
miguelgrinberg, line 356: This could be something I don't know, but I thought when running two queries as in this example you have to add RRF to combine the results into a single list. Is that optional? And if it is, how are the results from the two queries combined?

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
```diff
@@ -0,0 +1,562 @@
...
+     ]
+    }
+   ],
+   "source": [
```
miguelgrinberg, line 523:
```suggestion
   "source": [
    "query = \"What is biosimilarity?\"\n",
```

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
szabosteve: I don't think it's new. It's a [parameter of the `dense_vector` field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/dense-vector.html#dense-vector-params) that is available since 8.6.

miguelgrinberg (approved):

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
miguelgrinberg: Oh, okay, this is strange. I see it in the [typescript definition](https://github.com/elastic/elasticsearch-specification/blob/main/specification/_types/mapping/complex.ts#L52-L59), but for some reason it does not appear in the generated JSON spec. Need to figure out why.

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
szabosteve: `RRF` is optional and I _think_ the results are combined based on the `score` you get for the hits in case you didn't specify it. If we want to add `rrf` to the query, the `window_size` parameter needs to be tweaked to be greater than or equal to 100 (the `size` parameter) as its default value is `10`. I felt it would require some explanation that might make the notebook harder to comprehend. For this reason, I  decided not to add a specific ranking method.

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
szabosteve: Awesome, thank you for the thorough investigation @miguelgrinberg!

notebooks/integrations/cohere/cohere-elasticsearch.ipynb
miguelgrinberg: This is fine, I just did not know there was a default algorithm to combine the results.

PR #243, opened 2024-05-17
CI: remove warmup phase

maxjakob:
I don't remember clearly, but I have a feeling that we fixed an issue that we originally intended to solve with the warmup phase.

In this PR, we run CI without warmup X amount of times to show that things also work without it. This gives removes 2.5 to 5 minutes from the CI runs.


maxjakob:
ok the first run already failed 😅 

Though the failure doesn't seem related to the warmup in my opinion, I will stop here.

PR #246, opened 2024-05-21
Add per-version testing restrictions

miguelgrinberg:


.github/workflows/tests.yml
```diff
@@ -49,6 +49,6 @@ jobs:
         run: make install-nbtest
       - name: Warm up
         continue-on-error: true
-        run: sleep 30 && PATCH_ES=1 ELASTIC_CLOUD_ID=foo ELASTIC_API_KEY=bar bin/nbtest notebooks/search/00-quick-start.ipynb
+        run: sleep 30 && PATCH_ES=1 ELASTIC_CLOUD_ID=foo ELASTIC_API_KEY=bar ES_STACK=${{ matrix.es_stack }} bin/nbtest notebooks/search/00-quick-start.ipynb
```
joemcelroy, line 52: do we need to update the readme to add this env variable?

miguelgrinberg: Yes, I'll update the testing instructions in the `CONTRIBUTING.md` file.

miguelgrinberg:
@joemcelroy Okay, this is now looking pretty good.

- Updated the documentation in `CONTRIBUTING.md`
- Removed the pin we had on eland==8.12 (that we added because eland 8.13 does not work on 8.12 and older) and instead put the eland notebooks in the exempt list for 8.12. So now in 8.13 and 8.14 we are testing the latest eland version.

joemcelroy (approved):

PR #250, opened 2024-05-29
Adding notebook for function calling with Elasticsearch

ashishtiwari1993:
Adding a notebook for integrating OpenAI function calling with Elasticsearch


notebooks/integrations/openai/function-calling.ipynb
```diff
@@ -0,0 +1,502 @@
...
+    "\n",
+    "    return json_resp"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "6dc83930-89c0-465b-9002-e6aa0d3f5e6f",
+   "metadata": {},
+   "source": [
+    "### Generate Elasticsearch Query DSL based on user query"
```
joemcelroy, line 191: Worth looking into few-shot prompting so you can provide examples on the types of queries you want it to return. This should increase consistency.

notebooks/integrations/openai/function-calling.ipynb
```diff
@@ -0,0 +1,502 @@
...
+    "### Elastic\n",
+    "\n",
+    "Create an [Elastic Cloud deployment](https://www.elastic.co/search-labs/tutorials/install-elasticsearch/elastic-cloud) to get all Elastic credentials.  \n",
+    "`ES_API_KEY`: [Create](https://www.elastic.co/search-labs/tutorials/install-elasticsearch/elastic-cloud#creating-an-api-key) an API key.  \n",
+    "`ES_ENDPOINT`: [Copy](https://www.elastic.co/search-labs/tutorials/install-elasticsearch/elastic-cloud#finding-your-cloud-id) endpoint of Elasticsearch.\n",
+    "\n",
+    "### Open AI\n",
+    "\n",
+    "`OPENAI_API_KEY`: Setup an [Open AI account and create a secret key](https://platform.openai.com/docs/quickstart).  \n",
+    "`GPT_MODEL`: We’re going to use the `gpt-4-1106-vision-preview` model but you can check [here](https://platform.openai.com/docs/guides/function-calling) which model is being supported for function calling.\n",
```
joemcelroy, line 49: worth using gpt-4o instead?

joemcelroy (approved):

notebooks/integrations/openai/function-calling.ipynb
ashishtiwari1993: Hi Joe, I think adding this suggestion as side note would be better. Indeed its great suggestion but performing NL query is one of the part but focus is function calling. As I tested it working for most of the basic query i.e search, aggs.

notebooks/integrations/openai/function-calling.ipynb
ashishtiwari1993: Yeah I was about to change this. Updating with `gpt-4o`.

notebooks/integrations/openai/function-calling.ipynb
joemcelroy: sure! Likely it will work fine as is and yes mention in blog that we are relying on the model's own knowledge to generate the queries but we can increase reliability with few shot for more complex queries.

ashishtiwari1993: Good input. Thanks Joe

PR #271, opened 2024-06-17
Semantic text notebook

Mikep86:
Adds a search notebook demonstrating basic, end-to-end usage of the `semantic_text` field type


notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,527 @@
...
+   "id": "3db37d2cf8264468"
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "## Create Elastic Cloud deployment\n",
+    "\n",
+    "If you don't have an Elastic Cloud deployment, sign up [here](https://cloud.elastic.co/registration?utm_source=github&utm_content=elasticsearch-labs-notebook) for a free trial.\n",
+    "\n",
+    "TODO: Instruct user to disable ML node autoscaling?"
```
Mikep86, line 41: ML node autoscaling can make this notebook hard to use because it can take up to a few minutes for nodes to scale up from 0. Should we instruct users to disable ML node autoscaling and allocate at least one ML node to avoid this?

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,527 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# Semantic Search with Semantic Text\n",
+    "\n",
+    "<a target=\"_blank\" href=\"https://colab.research.google.com/github/Mikep86/elasticsearch-labs/blob/semantic-text-notebook/notebooks/search/09-semantic-text.ipynb\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>\n",
```
Mikep86, line 8: Will fix this link before merging

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,527 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "### Semantic Search with the `semantic` Query\n",
+    "\n",
+    "We can use the `semantic` query to quickly & easily query the `semantic_text` field in our index.\n",
+    "Under the hood, an embedding is automatically generated for our query text using the `semantic_text` field's inference endpoint.\n",
+    "\n",
+    "TODO: Add link to `semantic` query documentation"
```
Mikep86, line 391: Will fix once we know how documentation links should be handled (see [this comment](https://github.com/elastic/elasticsearch-labs/pull/271/files#r1642950911))

notebooks/search/09-semantic-text.ipynb
Mikep86, line 1: @miguelgrinberg Should I add cell output to this notebook?

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,510 @@
...
+   "cell_type": "markdown",
+   "source": [
+    "## Requirements\n",
+    "\n",
+    "For this example, you will need:\n",
+    "\n",
+    "- An Elastic deployment:\n",
+    "  - We'll be using [Elastic Cloud](https://www.elastic.co/guide/en/cloud/current/ec-getting-started.html) for this example (available with a [free trial](https://cloud.elastic.co/registration?utm_source=github&utm_content=elasticsearch-labs-notebook))\n",
+    "\n",
+    "- Elasticsearch 8.15 or above, or Elasticsearch serverless"
```
carlosdelest, line 27:
```suggestion
    "- Elasticsearch 8.15 or above, or [Elasticsearch serverless](https://www.elastic.co/elasticsearch/serverless)"
```

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,510 @@
...
+    "collapsed": false
+   },
+   "id": "6e14437dcce0f235"
+  },
+  {
+   "cell_type": "markdown",
+   "source": [
+    "### Enable Telemetry\n",
+    "\n",
+    "Knowing that you are using this notebook helps us decide where to invest our efforts to improve our products. We would like to ask you that you run the following code to let us gather anonymous usage statistics. See [telemetry.py](https://github.com/elastic/elasticsearch-labs/blob/main/telemetry/telemetry.py) for details. Thank you!"
```
carlosdelest, line 146:
```suggestion
    "Knowing that you are using this notebook helps us decide where to invest our efforts to improve our products, we would like to ask you that you run the following code to let us gather anonymous usage statistics. See [telemetry.py](https://github.com/elastic/elasticsearch-labs/blob/main/telemetry/telemetry.py) for details. Thank you!"
```

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,510 @@
...
+    "try:\n",
+    "    client.inference.delete_model(inference_id=\"my-elser-endpoint\")\n",
+    "except exceptions.NotFoundError:\n",
+    "    # Inference endpoint does not exist\n",
+    "    pass\n",
+    "\n",
+    "client.inference.put_model(\n",
+    "    task_type=\"sparse_embedding\",\n",
+    "    inference_id=\"my-elser-endpoint\",\n",
+    "    body={\n",
```
carlosdelest, line 233: I think we should ask request timeout here:

```suggestion
    "        \"request_timeout\"=60
    "    body={\n",
```

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,510 @@
...
+    "    index=\"semantic-text-movies\",\n",
+    "    query={\n",
+    "        \"bool\": {\n",
+    "            \"must\": {\n",
+    "                \"multi_match\": {\n",
+    "                    \"fields\": [\"plot\", \"genre\"],\n",
+    "                    \"query\": \"organized crime movies\",\n",
+    "                }\n",
+    "            },\n",
+    "            \"should\": {\n",
```
carlosdelest, line 447: Should we include boosting to let users know about linear combination of results?

carlosdelest (approved):
LGTM 👍  - some minor comments

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,527 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "source": [
+    "# Semantic Search with Semantic Text\n",
+    "\n",
+    "<a target=\"_blank\" href=\"https://colab.research.google.com/github/Mikep86/elasticsearch-labs/blob/semantic-text-notebook/notebooks/search/09-semantic-text.ipynb\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>\n",
+    "\n",
+    "Learn how to use the [semantic_text](https://www.elastic.co/guide/en/elasticsearch/reference/current/semantic-text.html) field type to quickly get started with semantic search."
```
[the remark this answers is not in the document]
leemthompo: @Mikep86 Look like you don't have much choice but to link to `master` initially. When 8.15 freezes you could then update the link to `8.15`.

notebooks/search/09-semantic-text.ipynb
miguelgrinberg: It's really up to you. If you want the CI to check that the output does not change then yes, you should. This may require you to add more mask regexes for things that are expected to be different in every run. See the `.nbtest.yml` file for the masks that are currently used in the notebooks in this directory.

miguelgrinberg:
@Mikep86 To prevent this notebook for running (and failing) in 8.14 and older CI runs, add a `EXEMPT_NOTEBOOKS__8_14` section in `bin/find-notebooks-to-run.sh` with this notebook and then that version and older will skip it.

notebooks/search/09-semantic-text.ipynb
joemcelroy: My suggestion is that you make the call to setup inference and you periodically check every 5 seconds to see the status of the inference status that it has been started. Once started, you can exit. This will allow the CI test suite to be run also. Example of this (old style which needs to be updated to use inference) https://github.com/elastic/elasticsearch-labs/blob/main/notebooks/search/03-ELSER.ipynb

notebooks/search/09-semantic-text.ipynb
Mikep86: This is boilerplate copy used on all the notebooks, so I think we should keep it uniform

notebooks/search/09-semantic-text.ipynb
Mikep86: Hmm, how far should we take that? Tuning linear combination is complex enough to deserve its own notebook. Adding a `boost` param to the example is simple, but doesn't do much to explain what we're doing and why. And on the other side of the spectrum, adding a whole section about linear combination steals focus from the main topic of this notebook.

Mikep86: Alternatively, we could turn the `multi_match` query into a filter to side-step all that potential complexity in this example.

carlosdelest: My concern is that users may not get good results for hybrid search. It might be good to show them the way for doing linear combination by exposing that particular knob to turn.

Agreed that this deserves its own blog post or link to documentation - just wanted to raise awareness for users that this is something they will need for hybrid search.

Mikep86: Would it be sufficient to add some example boosts and a short blurb about why those are there?

carlosdelest: For me, yes - nothing fancy, just letting users understand that need to be aware of linear combination for hybrid search 👍

notebooks/search/09-semantic-text.ipynb
```diff
@@ -0,0 +1,558 @@
...
+    "        print(\"ELSER Model is currently being deployed.\")\n",
+    "        continue\n",
+    "\n",
+    "    nodes = deployment_stats.get(\"nodes\")\n",
+    "    if nodes is not None and len(nodes) > 0:\n",
+    "        print(\"ELSER Model has been successfully deployed.\")\n",
+    "        break\n",
+    "    else:\n",
+    "        print(\"ELSER Model is currently being deployed.\")\n",
+    "    time.sleep(5)"
```
Mikep86, line 287: @joemcelroy I used the [ELSER notebook](https://github.com/elastic/elasticsearch-labs/blob/main/notebooks/search/03-ELSER.ipynb) as a guide for this code and modified it to work with inference endpoints. However, I found two issues when doing so:

- The notebook CI currently doesn't check against a serverless environment
- The [status check used in the ELSER notebook](https://github.com/elastic/elasticsearch-labs/blob/867785b2dcfb3232432c14050d6269fc00116846/notebooks/search/03-ELSER.ipynb#L263) breaks when run against a serverless environment. `trained_model_stats.0.deployment_stats.state` doesn't exist in the serverless response to `client.ml.get_trained_models_stats`. Instead, I checked if the `nodes` list is populated, which seems to work in both hosted & serverless environments.

notebooks/search/09-semantic-text.ipynb
Mikep86: Added, LMKWYT!

carlosdelest: Looks good, thanks! 💯

Mikep86:
@miguelgrinberg Seems like CI is broken on the `notebooks/langchain/langchain-using-own-model.ipynb` notebook

joemcelroy (approved):

PR #290, opened 2024-07-14
Supporting blog content/semantic text with amazon bedrock

llermaly:


joemcelroy:
need to run pre-commit. see contributing guide on how to do this.

PR #293, opened 2024-07-17
Change in supporting blog content

elastic-thanos:
* An implementation of confidence intervals with bootstrapping

### DIFF
--- a/supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
+++ b/supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
@@ -774,12 +774,54 @@
    ]
   },
   {
-   "cell_type": "code",
-   "execution_count": null,
-   "id": "0b804730-6969-4854-a56c-72e6b399f0a9",
    "metadata": {},
+   "cell_type": "markdown",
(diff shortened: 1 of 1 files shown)


supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
```diff
@@ -774,12 +774,54 @@
...
+   "id": "701ed82491660fd6"
+  },
+  {
    "metadata": {},
+   "cell_type": "code",
    "outputs": [],
-   "source": []
+   "execution_count": null,
+   "source": [
+    "ndcg_scores = post_reranking_eval_scores[\"ndcg_cut_10\"]\n",
```
tveasey, line 821: When I've used this library I always got the key named as "ndcg_cut@10" I'm sure you tested this but I wonder what the source of the discrepancy is (version?).

tveasey (approved):
LGTM

supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
elastic-thanos: Not sure about this. From what I see in `pytrec_eval`'s source code the supported formats are either `measure_k` or `measure.k`. Maybe it's a version issue as you say

PR #294, opened 2024-07-18
Add explanation for confidence intervals

elastic-thanos:
* A brief explanation and intro to confidence intervals and statistical significance

### DIFF
--- a/supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
+++ b/supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
@@ -790,13 +790,13 @@
    "metadata": {},
    "cell_type": "markdown",
    "source": [
-    "In this section we will briefly touch upon the concepts of confidence intervals and statistical significance.\n",
+    "In this section we will briefly touch upon the concepts of `confidence intervals` and `statistical significance` and we will see how we can use them to determine whether improvements in our pipelines are significant or not.\n",
     "\n",
     "We can think of it as follows: Our goal is to estimate the performance of our pipeline (retrieval and/or reranking) on a target corpus. Ideally, we would like to have access to **all** queries that our end-users will run against it but of course this is impossible. Instead, we have the set of test queries provided by the benchmark and we implicitly assume that the performance on this set can act as an accurate proxy of the overall performance (in the ideal scenario).\n",
     "\n",
-    "But we can make some extra assumptions to increase the reliability of our analysis. [Confidence intervals](https://en.wikipedia.org/wiki/Confidence_interval) , a concept from statistical theory, give us a tool to handle our uncertainty: By setting a certain level of confidence, let's go with 95% in this example, we can derive a range of values that will likely contain the parameter of interest (here the performance in the **ideal** scenario). In other words, if we repeated the same process an infinite number of times (by drawing different test sets) we could be confident that that in 95% of them the confidence interval would encompass the true value.\n",
+    "But we can make some extra assumptions to increase the reliability of our analysis. [Confidence intervals](https://en.wikipedia.org/wiki/Confidence_interval), a concept from statistical theory, give us a tool to handle our uncertainty. By setting a certain level of confidence, let's go with 95% in this example, we can derive a range of values that will likely contain the parameter of interest (here the performance in the **ideal** scenario). In other words, if we repeated the same process an infinite number of times (by drawing different test sets) we could be confident that in 95% of them the confidence interval would encompass the true value.\n",
     "\n",
-    "The code below shows an example of deriving confidence intervals using [bootstrapping](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)) combined with the `percentile` method. It should be noted that this statistic is affected a lot by the number and the variability of queries in the dataset i.e. smaller confidence intervals are expected for larger query sets and vice versa\n",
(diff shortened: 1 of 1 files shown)


supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
```diff
@@ -776,8 +777,29 @@
...
+    "\n",
+    "We can think of it as follows: Our goal is to estimate the performance of our pipeline (retrieval and/or reranking) on a target corpus. Ideally, we would like to have access to **all** queries that our end-users will run against it but of course this is impossible. Instead, we have the set of test queries provided by the benchmark and we implicitly assume that the performance on this set can act as an accurate proxy of the overall performance (in the ideal scenario).\n",
+    "\n",
+    "But we can make some extra assumptions to increase the reliability of our analysis. [Confidence intervals](https://en.wikipedia.org/wiki/Confidence_interval) , a concept from statistical theory, give us a tool to handle our uncertainty: By setting a certain level of confidence, let's go with 95% in this example, we can derive a range of values that will likely contain the parameter of interest (here the performance in the **ideal** scenario). In other words, if we repeated the same process an infinite number of times (by drawing different test sets) we could be confident that that in 95% of them the confidence interval would encompass the true value.\n",
```
tveasey, line 797:
```suggestion
    "But we can make some extra assumptions to increase the reliability of our analysis. [Confidence intervals](https://en.wikipedia.org/wiki/Confidence_interval), a concept from statistical theory, give us a tool to handle our uncertainty. By setting a certain level of confidence, let's go with 95% in this example, we can derive a range of values that will likely contain the parameter of interest (here the performance in the **ideal** scenario). In other words, if we repeated the same process an infinite number of times (by drawing different test sets) we could be confident that that in 95% of them the confidence interval would encompass the true value.\n",
```

maxjakob:
```suggestion
    "But we can make some extra assumptions to increase the reliability of our analysis. [Confidence intervals](https://en.wikipedia.org/wiki/Confidence_interval), a concept from statistical theory, give us a tool to handle our uncertainty. By setting a certain level of confidence, let's go with 95% in this example, we can derive a range of values that will likely contain the parameter of interest (here the performance in the **ideal** scenario). In other words, if we repeated the same process an infinite number of times (by drawing different test sets) we could be confident that in 95% of them the confidence interval would encompass the true value.\n",
```

Tom's suggestion and deduplicating `that`.

supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
```diff
@@ -776,8 +777,29 @@
...
+    "\n",
+    "The code below shows an example of deriving confidence intervals using [bootstrapping](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)) combined with the `percentile` method. It should be noted that this statistic is affected a lot by the number and the variability of queries in the dataset i.e. smaller confidence intervals are expected for larger query sets and vice versa\n",
```
maxjakob, line 799:
```suggestion
    "The code below shows an example of deriving confidence intervals using [bootstrapping](https://en.wikipedia.org/wiki/Bootstrapping_(statistics) combined with the `percentile` method. It should be noted that this statistic is affected a lot by the number and the variability of queries in the dataset i.e. smaller confidence intervals are expected for larger query sets and vice versa\n",
```

supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
```diff
@@ -776,8 +777,29 @@
...
+   "metadata": {},
+   "cell_type": "markdown",
+   "source": "### Confidence intervals",
+   "id": "392ab084d5bc024"
+  },
+  {
+   "metadata": {},
+   "cell_type": "markdown",
+   "source": [
+    "In this section we will briefly touch upon the concepts of confidence intervals and statistical significance.\n",
```
maxjakob, line 793:
```suggestion
    "In this section we will briefly touch upon the concepts of confidence intervals and statistical significance. We can use these to determine if a retrieval or reranking improvement is significant or can be considered negligible.\n",
```

maxjakob (approved):
Nice contextualization!

supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
elastic-thanos: Here, the right parenthesis `)` is part of Wikipedia's URL. I know it's a bit unfortunate :)

maxjakob: ah, then we might have to escape it, since not it renders like this:
image

elastic-thanos: Right. I will do that 👍

supporting-blog-content/evaluating-search-relevance-part-1/retrieve-and-rerank.ipynb
elastic-thanos: I slightly adjusted the phrasing

PR #332, opened 2024-09-23
Upgrade chatbot example to langchain 0.3

miguelgrinberg:
I have updated the chatbot RAG example to use the recently released langchain 0.3.

### DIFF
--- a/example-apps/chatbot-rag-app/requirements.in
+++ b/example-apps/chatbot-rag-app/requirements.in
@@ -4,6 +4,7 @@ pip-tools
 # core dependencies
 elasticsearch
 langchain
+langchain-community
 langchain-elasticsearch
 tiktoken
 flask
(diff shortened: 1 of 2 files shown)


alizenhom:
does the chatbot rag example work without this PR?

i have tried to spin up the chatbot RAG app, followed the documentation step by step but there is a dependency conflict happening


i always get this error when trying to run `flask run`

`ModuleNotFoundError: No module named 'langchain_elasticsearch'`

also `langchain-community` is missing from dependecies

miguelgrinberg:
I believe the example is currently broken, at least I got a report of dependency issues, which is why I looked into this.

The langchain-community package used to be part of the core langchain, and it isn't anymore. I have added it as a top-level dependency in this PR.

I'm not sure why you get errors for langchain-elasticsearch. This package was also split from core langchain a while ago, but it is explicitly listed in the requirements already.

joemcelroy (approved):

PR #333, opened 2024-09-25
Remove unnecessary argument

elastic-thanos:


pquentin (approved):
Thanks! LGTM.

For what it's worth, we discussed this offline with @elastic-thanos:

 * The correct way to refresh bulk calls is to add `refresh="wait_for"` to the bulk() call. The existing way was only adding a `refresh` field in the documents
 * That said, doing that will refresh for every single bulk call (by default every 500 docs) and it is more efficient to only do it once at the end, which is already done in this notebook anyway.

PR #348, opened 2024-11-01
Update App Search Migration Notebook to Correct Some Errors and Warnings, and Add More Descriptive Text

markjhoy:
Updates the App Search migration notebook for the following:

* Prints the synonym set created after it is created
* Uses `put_ruleset` instead of `put` when adding the query rules (was throwing an error with the latest Elasticsearch client)
* Breaks the index creation piece up into smaller code chunks, and adds descriptive text for what each part does
* Adds a bit more clarity around the ELSER model name
* Adds more description around the App Search query; how it's built and what the sections do.

### DIFF
--- a/notebooks/enterprise-search/app-search-engine-exporter.ipynb
+++ b/notebooks/enterprise-search/app-search-engine-exporter.ipynb
@@ -606,10 +606,10 @@
     "        \"service_settings\": {\n",
     "            \"model_id\": \".elser_model_2_linux-x86_64\",\n",
     "            \"num_allocations\": 1,\n",
-    "            \"num_threads\": 1\n",
+    "            \"num_threads\": 1,\n",
     "        }\n",
     "    },\n",
-    "    task_type=\"sparse_embedding\"\n",
+    "    task_type=\"sparse_embedding\",\n",
     ")\n"
    ]
   },
@@ -645,7 +645,7 @@
     "# and for our text fields, add a \"copy_to\" directive to copy the text to the semantic_text field\n",
     "for field_name in SEMANTIC_TEXT_FIELDS:\n",
     "    semantic_field_name = field_name + \"_semantic\"\n",
-    "    mapping[field_name].update({ \"copy_to\": semantic_field_name })\n",
+    "    mapping[field_name].update({\"copy_to\": semantic_field_name})\n",
     "\n",
     "elasticsearch.indices.put_mapping(index=DEST_INDEX, properties=mapping)"
    ]
@@ -950,9 +950,7 @@
     "payload = app_search_query_payload.copy()\n",
     "\n",
     "for semantic_text_query in semantic_text_queries:\n",
-    "    payload[\"query\"][\"rule\"][\"organic\"][\"bool\"][\"should\"].append(\n",
-    "        semantic_text_query\n",
-    "    )\n",
+    "    payload[\"query\"][\"rule\"][\"organic\"][\"bool\"][\"should\"].append(semantic_text_query)\n",
     "\n",
     "print(f\"Elasticsearch payload:\\n{json.dumps(payload, indent=2)}\\n\")"
    ]
--- a/notebooks/enterprise-search/app-search-engine-exporter.ipynb
+++ b/notebooks/enterprise-search/app-search-engine-exporter.ipynb
@@ -579,7 +579,7 @@
    "source": [
     "# Add semantic text fields for semantic search (optional)\n",
     "\n",
-    "One of the advantages of having our exported index directly in Elasticsearch is that we can easily take advantage of doing semantic search with ELSER. To do this, we'll need to add an inference endpoint using ELSER, and a `semantic_text` field to our index to use it.\n",
+    "One of the advantages of exporting our index directly to Elasticsearch is that we can easily perform semantic search with ELSER. To do this, we'll need to add an inference endpoint using ELSER, and a `semantic_text` field to our index to use it.\n",
     "\n",
     "Note that to use this feature, your cluster must have at least one ML node set up with enough resources allocated to it.\n",
     "\n",
@@ -655,7 +655,7 @@
    "metadata": {},
    "source": [
     "## Reindex the data\n",
-    "Now that we have created the Elasticsearch index, it's time to reindex our data in the new index. If you are using the `semantic_text` fields as defined above with a `_semantic` suffix, and then the reindexing process with automatically infer the sparse vector values from ELSER and use those for the vectors as the reindex takes place."
+    "Now that we have created the Elasticsearch index, it's time to reindex our data in the new index. If you are using the `semantic_text` fields as defined above with a `_semantic` suffix, and then the reindexing process will automatically infer the sparse vector values from ELSER and use those for the vectors as the reindex takes place."
    ]
   },
   {
@@ -816,7 +816,7 @@
     "                    ]\n",
     "                }\n",
     "            },\n",
-    "            \"ruleset_ids\": ENGINE_NAME,\n",
+    "            \"ruleset_ids\": [ENGINE_NAME],\n",
     "            \"match_criteria\": {\"user_query\": QUERY_STRING},\n",
     "        }\n",
     "    },\n",
@@ -910,7 +910,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### How to combine App Search queries with Semantic Text\n",
+    "### How to combine App Search queries with semantic text\n",
     "\n",
     "We will now provide an example on how to combine the previous two queries into a single query that applies both BM25 search and semantic search.\n",
     "In the previous examples, we have a `bool` query with `should` clauses.\n",
--- a/notebooks/enterprise-search/app-search-engine-exporter.ipynb
+++ b/notebooks/enterprise-search/app-search-engine-exporter.ipynb
@@ -66,7 +66,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -100,7 +100,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -129,7 +129,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 10,
    "metadata": {
     "id": "kpV8K5jHvRK6"
    },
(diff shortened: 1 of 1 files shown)


kderusso (commented):
The changes look good! 

The high level feedback I have though, outside of some of the specific feedback that I gave, is that we should consider pointing people to retrievers in these examples. This will be especially important for the rule query which will not work with rerankers (but the rule retriever will work with rerank retrievers such as RRF or semantic reranking). It wouldn't be too hard to migrate these from the query DSL to retriever syntax and it has the added advantage that it will be consistent with what we do in Playground. Thanks!

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -147,6 +147,22 @@
...
+    "Let's take a quick look at the synonyms we've migrated. We'll do this via the `GET _synonyms` endpoint"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(json.dumps(elasticsearch.synonyms.get_synonym(id=ENGINE_NAME).body, indent=2))"
```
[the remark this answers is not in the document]
markjhoy: I would tend to agree - but the client has it as `get_synonym` ... :/

image

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -277,8 +299,210 @@
...
     "if elasticsearch.indices.exists(index=DEST_INDEX):\n",
-    "    elasticsearch.indices.delete(index=DEST_INDEX)\n",
+    "    elasticsearch.indices.delete(index=DEST_INDEX)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Next, we'll create our settings which includes filters and and analyzers to use for our text fields.\n",
```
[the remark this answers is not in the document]
markjhoy: Yep - I have a forthcoming blog post that I was planning on diving into this a bit more as well.

notebooks/enterprise-search/app-search-engine-exporter.ipynb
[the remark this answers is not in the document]
kderusso: Yeah I ran through the notebook to verify, I would definitely say it's a bug in the specfication.

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -711,7 +882,9 @@
     "\n",
     "If you [enabled and reindexed your data with ELSER](#add-sparse_vector-fields-for-semantic-search-optional), we can now use this to do semantic search.\n",
     "For each `spare_vector` we will generate a `text_expansion` query. These `text_expansion` queries will be added as `should` clauses to a top-level `bool` query.\n",
-    "We also use `min_score` because we want to exclude less relevant results. "
+    "We also use `min_score` because we want to exclude less relevant results. \n",
```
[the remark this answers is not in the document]
ioanatia: it's all good on my side if @markjhoy wants to follow up on this in a follow up PR or do it here.
changing from `text_expansion` to use `semantic` will require changing the section on how we index these fields as well - right now the notebook has a guide on how to use ingest pipelines to use ELSER with `text_expansion`.
Using `semantic_text` instead would be much better - we can also point users to use E5 when dealing with a non-english dataset.

ioanatia (approved):
I think the changes look good, provided we also address the feedback from Kathleen.
We can make the changes to move from ingest pipelines with ELSER to semantic_text in a follow up PR of here - it's up to you.

I have not looked into the CI failures whether they are related.

markjhoy:
> I have not looked into the CI failures whether they are related.

Ugh - it looks like the CI failures are version specific issues... I'll have to see if I can either (a) lock the notebook to 8.16+ - or, make the code version specific...

markjhoy:
> > I have not looked into the CI failures whether they are related.
> 
> Ugh - it looks like the CI failures are version specific issues... I'll have to see if I can either (a) lock the notebook to 8.16+ - or, make the code version specific for backwards compatibility...

Oh wait - nevermind this... the errors we are seeing in the CI are actually from other notebooks, namely:
* `notebooks/document-chunking/with-index-pipelines.ipynb... Failed`
* `notebooks/document-chunking/with-langchain-splitters.ipynb... Failed`
* `notebooks/integrations/hugging-face/loading-model-from-hugging-face.ipynb... Failed`
* `notebooks/langchain/langchain-using-own-model.ipynb... Failed`

It appears that our migration notebook is fine...

markjhoy:
I've updated the notebook to use `semantic_text` search. That new field type makes it so easy to do semantic search now.

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -455,13 +577,15 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Add `sparse_vector` fields for semantic search (optional)\n",
+    "# Add semantic text fields for semantic search (optional)\n",
     "\n",
-    "One of the advantages of having our exported index directly in Elasticsearch is that we can easily take advantage of doing semantic search with ELSER. To do this, we'll need to add a `sparse_vector` field to our index, set up an ingest pipeline, and reindex our data.\n",
+    "One of the advantages of having our exported index directly in Elasticsearch is that we can easily take advantage of doing semantic search with ELSER. To do this, we'll need to add an inference endpoint using ELSER, and a `semantic_text` field to our index to use it.\n",
```
kderusso, line 582: Nit: Trying to avoid two uses of `advantage` in the same sentence

```suggestion
    "One of the advantages of exporting our index directly to Elasticsearch is that we can easily perform semantic search with ELSER. To do this, we'll need to add an inference endpoint using ELSER, and a `semantic_text` field to our index to use it.\n",
```

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -470,29 +594,34 @@
...
+    "if elasticsearch.inference.get(inference_id=\"elser_inference_endpoint\"):\n",
+    "    elasticsearch.inference.delete(inference_id=\"elser_inference_endpoint\")\n",
+    "\n",
+    "# and create our endpoint using the ELSER v2 model\n",
+    "elasticsearch.inference.put(\n",
+    "    inference_id='elser_inference_endpoint',\n",
+    "    inference_config={\n",
+    "        \"service\": \"elasticsearch\",\n",
+    "        \"service_settings\": {\n",
+    "            \"model_id\": \".elser_model_2_linux-x86_64\",\n",
```
kderusso, line 607: Should we add a note that this will not work on aarch64? Or is that moot because we're using cloud for the examples?

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -501,62 +630,43 @@
...
+    "elasticsearch.indices.put_mapping(index=DEST_INDEX, properties=mapping)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
     "## Reindex the data\n",
-    "Now that we have created the Elasticsearch index and the ingest pipeline, it's time to reindex our data in the new index. The pipeline definition we created above will create a field for each of the `SPARSE_VECTOR_FIELDS` we defined with a `_semantic` suffix, and then infer the sparse vector values from ELSER as the reindex takes place."
+    "Now that we have created the Elasticsearch index, it's time to reindex our data in the new index. If you are using the `semantic_text` fields as defined above with a `_semantic` suffix, and then the reindexing process with automatically infer the sparse vector values from ELSER and use those for the vectors as the reindex takes place."
```
kderusso, line 658:
```suggestion
    "Now that we have created the Elasticsearch index, it's time to reindex our data in the new index. If you are using the `semantic_text` fields as defined above with a `_semantic` suffix, and then the reindexing process will automatically infer the sparse vector values from ELSER and use those for the vectors as the reindex takes place."
```

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -658,7 +816,7 @@
     "                    ]\n",
     "                }\n",
     "            },\n",
-    "            \"ruleset_id\": ENGINE_NAME,\n",
+    "            \"ruleset_ids\": ENGINE_NAME,\n",
```
kderusso, line 819: `ruleset_ids` needs to be an array of strings

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -743,15 +902,15 @@
...
+    "results = elasticsearch.search(index=DEST_INDEX, query=semantic_query, min_score=1)\n",
     "print(f\"Query results:\\n{json.dumps(results.body, indent=2)}\\n\")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### How to combine App Search queries with ELSER\n",
+    "### How to combine App Search queries with Semantic Text\n",
```
kderusso, line 913: Do we want this to be capitalized?

notebooks/enterprise-search/app-search-engine-exporter.ipynb
```diff
@@ -620,20 +735,63 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "\n",
     "app_search_query_payload = {\n",
     "    \"query\": {\n",
-    "        \"rule_query\": {\n",
+    "        \"rule\": {\n",
```
kderusso, line 794: It would have been nice to use retrievers here, perhaps we could wrap this query under a standard retriever (or a rule retriever!) or even use RRF?

kderusso (approved):
Looks good, thanks for iterating! 

I left some feedback, but it is non blocking. I do think we should consider being retrievers-first here, but I defer to you and the ent search deprecation team on the final decision.

notebooks/enterprise-search/app-search-engine-exporter.ipynb
markjhoy: Yeah - since it's cloud - it's a moot point.

markjhoy:
> I left some feedback, but it is non blocking. I do think we should consider being retrievers-first here, but I defer to you and the ent search deprecation team on the final decision.

Thanks @kderusso! Yeah - for the time being, I'd like to get this version out now, and come back and do a pass using retrievers. I need to still brush up a bit more myself on them but at least using this version with the semantic text fields is a solid upgrade.

markjhoy:
The failures in the CI are from other notebooks and not this (isolated) one. Going to merge this.

PR #362, opened 2024-12-19
Updates chatbot-rag-app and simplifies instructions

codefromthecrypt:
This updates chatbot-rag-app to the latest libraries and simplifies instructions by delegating details of configuration to the .env file.

This also works around problems we found in DEBUG mode, by using production mode regardless of task. Doing so notably will allow tracing to work in the future.

Efforts here were thanks to many, despite me raising it, notably @joemcelroy @ezimuel @xrmx and @anuraaga


example-apps/chatbot-rag-app/README.md
```diff
@@ -15,216 +15,111 @@ curl https://codeload.github.com/elastic/elasticsearch-labs/tar.gz/main | \
...
+Model downloaded, starting deployment
+Loading data from ./data/data.json
+Loaded 15 documents
+Split 15 documents into 26 chunks
+Creating Elasticsearch sparse vector store in http://localhost:9200
 ```
 
-By default, this will index the data into the `workplace-app-docs` index. You can change this by setting the `ES_INDEX` environment variable.
-
-##### Indexing your own data
```
codefromthecrypt, line 212: @joemcelroy so I think this since it applies regardless of how you run it.. Maybe should be in a different section, like `## Customizing the data` that is if we want to encourage them to do it ;)

example-apps/chatbot-rag-app/README.md
```diff
@@ -15,216 +15,111 @@ curl https://codeload.github.com/elastic/elasticsearch-labs/tar.gz/main | \
...
+*Note*: First time creating the index can fail on timeout. Wait a few minutes
+and retry.
 
-Make sure you have a `.env` file with all your variables, then run:
+### Run locally
 
-```sh
-docker run --rm --env-file .env chatbot-rag-app flask create-index
-```
+If you want to run this example with Python and Node.js, you need to do a few
```
anuraaga, line 71: Should we call out these instructions are specifically for running locally in production mode?

codefromthecrypt: yeah because if we use DEBUG mode then later add opentelemetry things will break ;)

example-apps/chatbot-rag-app/README.md
codefromthecrypt: added back at the bottom, fixing dead links

joemcelroy:
Changes look great overall - one small request please:

Could we focus on connecting to Elasticsearch in three scenarios:
   - url only (for localhost:9200, no auth)
   - url with user / pass
   - url with api-key

Drop connecting with cloud id, we have been moving away from this for a while and unnecessary complicates the connection logic.

joemcelroy (approved):

anuraaga (approved):

PR #371, opened 2025-01-09
chatbot-rag-app: adds openteletry for cohere with langtrace

codefromthecrypt:
This adds langtrace to add more LLM support, starting with cohere. Thanks to @xrmx for the help with this.

After this change, if you follow our README for opentelemetry, you'll see a GenAI span for Cohere despite it not yet supported in [EDOT Python](https://github.com/elastic/elastic-otel-python).

Screenshot 2025-01-09 at 1 35 08 PM

It is understood that other LLM platforms are desired, notably bedrock. Let's merge this first so that @karthikscale3 or others from Langtrace can add more. Ideally, configuration for each LLM will be added in separate PR, so that it is easier to tell what changes were needed.

Closes #357


xrmx (approved):

xrmx (changes requested):
@codefromthecrypt I think this is missing starting the instrumentation when cohere is set as llm type

xrmx (approved):

codefromthecrypt:
yep thanks for the catch @xrmx

PR #378, opened 2025-01-17
re-uses docker setup in chatbot-rag-app and openai-embeddings examples

codefromthecrypt:
This removes copy/paste of the same docker-compose-elastic.yml file between two examples.


example-apps/openai-embeddings/README.md
```diff
@@ -144,3 +147,6 @@ EDOT Node.js [here](https://github.com/elastic/elastic-otel-node).
 
 - Modify the initialization of `elasticsearchClient` in [utils.js](utils.js)
 - Refer to [this document](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-connecting.html)
+
+---
+[install-es]: https://www.elastic.co/search-labs/tutorials/install-elasticsearch
```
joemcelroy, line 152: @leemthompo is there a better link for this now? We should drop this page ago (the docker one anyway with start-local or a elastic doc page).

leemthompo: or if you want overview of different deployment options, use:

https://www.elastic.co/guide/en/elasticsearch/reference/current/elasticsearch-intro-deploy.html

xrmx: We need the APM server that is not considered in any of the elasticsearch doc AFAICS

xrmx (approved):

joemcelroy (approved):

example-apps/openai-embeddings/README.md
codefromthecrypt: agree to take it off here for now. Maybe folks can help prioritize https://github.com/elastic/start-local-private/issues/58 which would be the best link when solved.

PR #388, opened 2025-01-30
Copy update to instructions for clarity

mikeisonthenets:
Based on feedback from non-native English-speaking customers relying heavily on App Search in production, I've updated this text to be more explicit. The previous wording confused whether reindexing was optional during migration, so I've clarified the strong recommendation to reindex data. Also added some code markdown to make data types stand out.


markjhoy (approved):
Looks good. Thanks!

PR #395, opened 2025-02-13
Use published images in chatbot-rag-app example

anuraaga:
Adds the published image to the chatbot-rag-app docker-compose. It's used by default if `--build` isn't passed.

### DIFF
--- a/example-apps/chatbot-rag-app/README.md
+++ b/example-apps/chatbot-rag-app/README.md
@@ -57,7 +57,7 @@ Docker compose is the easiest way, as you get one-step to:
 **Double-check you have a `.env` file with all your variables set first!**
 
 ```bash
-docker compose up --force-recreate
+docker compose up --pull always --force-recreate
 ```
 
 *Note*: First time creating the index can fail on timeout. Wait a few minutes
--- a/example-apps/chatbot-rag-app/README.md
+++ b/example-apps/chatbot-rag-app/README.md
@@ -51,14 +51,13 @@ ease while locally is advised if you are making changes to the application.
 ### Run with docker
 
 Docker compose is the easiest way, as you get one-step to:
-* build the [frontend](frontend)
 * ingest data into elasticsearch
 * run the app, which listens on http://localhost:4000
 
 **Double-check you have a `.env` file with all your variables set first!**
 
 ```bash
-docker compose up --build --force-recreate
+docker compose up --force-recreate
 ```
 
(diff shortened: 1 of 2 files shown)


example-apps/chatbot-rag-app/README.md
```diff
@@ -51,14 +51,13 @@ ease while locally is advised if you are making changes to the application.
...
 Docker compose is the easiest way, as you get one-step to:
-* build the [frontend](frontend)
 * ingest data into elasticsearch
 * run the app, which listens on http://localhost:4000
 
 **Double-check you have a `.env` file with all your variables set first!**
 
 ```bash
-docker compose up --build --force-recreate
+docker compose up --force-recreate
```
anuraaga, line 60: I considered `--pull always` for reproducibility but figured the simpler command is better

anuraaga (commented):
@codefromthecrypt** Can you take a look?

example-apps/chatbot-rag-app/README.md
codefromthecrypt: agreed in principal, but `--force-recreate` is already there, too. I think we should do `--pull-always` until the telemetry stuff settles down, so that a future blog won't result in someone retrying and not seeing the new stuff. I know it may be overly cautious, so not hard opinion on this.

anuraaga: Yeah given possible telemetry churn makes sense, added it

example-apps/chatbot-rag-app/README.md
```diff
@@ -186,6 +185,14 @@ passages. Modify this script to index your own data.
 
 See [Langchain documentation][loader-docs] for more ways to load documents.
 
+### Building from source with docker
+
+To build the app with Docker instead of using published images, pass the `--build`
```
codefromthecrypt, line 190:
```suggestion
To build the app from source instead of using published images, pass the `--build`
```

example-apps/chatbot-rag-app/README.md
```diff
@@ -186,6 +185,14 @@ passages. Modify this script to index your own data.
...
+flag to Compose.
```
codefromthecrypt, line 191:
```suggestion
flag to Docker Compose.
```

codefromthecrypt (approved):
editorial only. LGTM

codefromthecrypt:
thx again!

Issue #379, opened 2025-01-17
chatbot-rag-app: re-introduce ES cloud configuration


codefromthecrypt:
When we refactored chatbot-rag-app, we started with local ENV. That said, many folks use ES cloud. We should add support in code and env.example for ES cloud connections, including opentelemetry setup.

Ideally, this can use canonical ENV variables used in other elastic tools, without us having to hard code if statements about cloud vs not. That said, I'm not sure if elasticsearch-py supports that at the moment.

xeraa:
Why does this have to be Cloud specific? Won't the ELASTICSEARCH_URL (or whatever you're using) work locally as well as in any other environment?

codefromthecrypt:
@xeraa Agree on the main app. This is more about the opentelemetry part, to make README or env.example comments of what to do when you are using Elastic cloud (for APM).

codefromthecrypt:
Notes for when this is implemented cc @bshetti This comes after k8s in #396, and needs to do the following

- [ ] Update env.example directions ELASTICSEARCH_URL such as we have for k8s, with an example cloud URL saying you also need to set ELASTICSEARCH_API_KEY
- [ ] Update README with a section about Elastic Cloud and make sure it is an anchor so we can reference it in the three ways to run: Docker, Python and Kubernetes
- [ ] Test default where OTEL_EXPORTER_OTLP_ENDPOINT is ignored due to default `OTEL_SDK_DISABLED=true` on Docker, Python and Kubernetes. So this means setting ELASTICSEARCH_URL and ELASTICSEARCH_API_KEY

- [ ] Update env.example directions OTEL_EXPORTER_OTLP_ENDPOINT such as we have for k8s, also with authentication instructions for that.
- [ ] Update README opentelemetry section about elastic cloud, saying anything notable about how to get auth working
- [ ] Test OTEL_EXPORTER_OTLP_ENDPOINT when `OTEL_SDK_DISABLED=false` on Docker, Python and Kubernetes. 

So, this is why I'm asking this to be solved in a separate PR as if this isn't done coherently someone will have to go back and clean it up later, or it would get lost in infrastructure for other changes. Hope it makes sense.

codefromthecrypt:
k8s is merged, so if anyone wants this done elasticon week, it's on them as I am busy literally every day. if you do, many thanks!

Issue #307, opened 2024-08-06
chatbot-rag-app: timeout errors installing elser and/or while it is initializing


epugh:
I noticed that with my under powered laptop, when I run ES in Docker, and run the `flask create-index` i get a time out the first time, but it works teh second time:

```
(.venv) ➜  chatbot-rag-app git:(main) ✗ flask create-index
".elser_model_2" model not available, downloading it now
Model downloaded, starting deployment
Traceback (most recent call last):
  File "/Users/epugh/Documents/clients/OSC/rag/elasticsearch-labs/example-apps/chatbot-rag-app/.venv/bin/flask", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/Users/epugh/Documents/clients/OSC/rag/elasticsearch-labs/example-apps/chatbot-rag-app/.venv/lib/python3.11/site-packages/flask/cli.py", line 1064, in main
    cli.main()
  File "/Users/epugh/Documents/clients/OSC/rag/elasticsearch-labs/example-apps/chatbot-rag-app/.venv/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/epugh/Documents/clients/OSC/rag/elasticsearch-labs/example-apps/chatbot-rag-app/.venv/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/epugh/Documents/clients/OSC/rag/elasticsearch-labs/example-apps/chatbot-rag-app/.venv/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/epugh/Documents/clients/OSC/rag/elasticsearch-labs/example-apps/chatbot-rag-app/.venv/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #396, opened 2025-02-17
chatbot-rag-app: adds Kubernetes manifest and instructions

### LINKED ISSUE #366: chatbot-rag-app: add instructions for k8s deployment
Right now, there are instructions for chatbot-rag-app docker and python, but not k8s. @bshetti was interested in this.

k8s is more complicated as it requires a build step implicit when using docker-compose
  (e.g. docker compose --build means you don't have to add a separate build image step)
Also, it needs to take note of all ENV variables set otherwise via .env in existing instructions

It may be possible that some layered tool over k8s can simplify this to the same easy experience as docker compose.

Meanwhile, folks should use normal k8s port forwarding as the example itself can access ES via localhost if the port is forwarded and doesn't need to be deployed inside k8s.

https://github.com/elastic/elasticsearch-labs/tree/main/example-apps/chatbot-rag-app

codefromthecrypt:
Decided to action this so that we have a coherent experience between docker compose and k8s. This is as close as I could get it. If folks have feedback or a different direction, do tell!

Fixes #366

### DIFF
--- a/example-apps/chatbot-rag-app/env.example
+++ b/example-apps/chatbot-rag-app/env.example
@@ -6,6 +6,8 @@ FLASK_APP=api/app.py
 PYTHONUNBUFFERED=1
 
 # How you connect to Elasticsearch: change details to your instance
+# This defaults to a Elastic Stack accessible via localhost. When this is
+# running inside Kubernetes, update to http://elasticsearch:9200 or similar.
 ELASTICSEARCH_URL=http://localhost:9200
 ELASTICSEARCH_USER=elastic
 ELASTICSEARCH_PASSWORD=elastic
@@ -68,7 +70,9 @@ OTEL_SDK_DISABLED=true
 # Assign the service name that shows up in Kibana
 OTEL_SERVICE_NAME=chatbot-rag-app
 
-# Default to send traces to the Elastic APM server
+# Default to send logs, traces and metrics to an Elastic APM server accessible
+# via localhost. If using running inside k8s, update to http://apm-server:8200
+# or similar.
 OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:8200
 OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
 
--- a/example-apps/chatbot-rag-app/env.example
+++ b/example-apps/chatbot-rag-app/env.example
@@ -6,8 +6,10 @@ FLASK_APP=api/app.py
 PYTHONUNBUFFERED=1
 
 # How you connect to Elasticsearch: change details to your instance
-# This defaults to a Elastic Stack accessible via localhost. When this is
-# running inside Kubernetes, update to http://elasticsearch:9200 or similar.
+# This defaults to a Elastic Stack accessible via localhost.
+#
+# When running inside Kubernetes, set to http://elasticsearch.default.svc:9200
+# or similar.
 ELASTICSEARCH_URL=http://localhost:9200
 ELASTICSEARCH_USER=elastic
 ELASTICSEARCH_PASSWORD=elastic
@@ -71,7 +73,9 @@ OTEL_SDK_DISABLED=true
 OTEL_SERVICE_NAME=chatbot-rag-app
 
 # Default to send logs, traces and metrics to an Elastic APM server accessible
-# via localhost. If using running inside k8s, update to http://apm-server:8200
+# via localhost.
+#
+# When running inside Kubernetes, set to http://elasticsearch.default.svc:9200
 # or similar.
 OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:8200
 OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
--- a/example-apps/chatbot-rag-app/k8s-manifest.yml
+++ b/example-apps/chatbot-rag-app/k8s-manifest.yml
@@ -1,4 +1,6 @@
 ---
+# chatbot-rag-app deploys "create-index" to install ELSER and load values.
+# Then, it starts "api-frontend" to serve the application.
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -18,29 +20,30 @@ spec:
         - name: gcloud-credentials
           secret:
             secretName: gcloud-credentials
+            optional: true  # only read when `LLM_TYPE=vertex`
       initContainers:
-      - name: create-index
-        image: &image ghcr.io/elastic/elasticsearch-labs/chatbot-rag-app:latest
-        command: &command [ "bash", "-eu", "./entrypoint.sh" ] # match image
-        args: ["flask", "create-index"]
-        # This recreates your configmap based on your .env file:
-        # kubectl create configmap chatbot-rag-app-env --from-env-file=.env
-        envFrom: &envFrom
-          - configMapRef:
-              name: chatbot-rag-app-env
+        - name: create-index
+          image: &image ghcr.io/elastic/elasticsearch-labs/chatbot-rag-app:latest
+          command: &command [ "bash", "-eu", "./entrypoint.sh" ] # match image
+          args: [ "flask", "create-index" ]
+          # This recreates your configmap based on your .env file:
+          # kubectl create configmap chatbot-rag-app-env --from-env-file=.env
+          envFrom: &envFrom
+            - configMapRef:
+                name: chatbot-rag-app-env
       containers:
-      - name: api-frontend
-        image: *image
-        command: *command
-        args: [ "python", "api/app.py" ]
-        ports:
-          - containerPort: 4000
-        envFrom: *envFrom
-        # For `LLM_TYPE=vertex`: mount credentials to the path read by the google-cloud-sdk
-        volumeMounts:
-          - name: gcloud-credentials
-            mountPath: /root/.config/gcloud
-            readOnly: true
+        - name: api-frontend
+          image: *image
+          command: *command
+          args: [ "python", "api/app.py" ]
+          ports:
+            - containerPort: 4000
+          envFrom: *envFrom
+          # For `LLM_TYPE=vertex`: mount credentials to the path read by the google-cloud-sdk
+          volumeMounts:
+            - name: gcloud-credentials
+              mountPath: /root/.config/gcloud
+              readOnly: true
 ---
 apiVersion: v1
 kind: Service
--- a/example-apps/chatbot-rag-app/k8s-manifest.yml
+++ b/example-apps/chatbot-rag-app/k8s-manifest.yml
@@ -13,9 +13,7 @@ spec:
       labels:
         app: chatbot-rag-app
     spec:
-      # The below will recreate your secret based on the gcloud credentials file
-      # kubectl create secret generic gcloud-credentials \
-      #  --from-file=application_default_credentials.json=$HOME/.config/gcloud/application_default_credentials.json
+      # For `LLM_TYPE=vertex`: create a volume for application_default_credentials.json
       volumes:
         - name: gcloud-credentials
           secret:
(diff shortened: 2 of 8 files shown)


codefromthecrypt:
note: each thing we do runs back into this. it would be great to have a way to quickly initialize elser not just installing it, but first time use without timeouts for several minutes https://github.com/elastic/elasticsearch-labs/issues/307

example-apps/chatbot-rag-app/k8s-manifest.yml
```diff
@@ -0,0 +1,61 @@
...
+  selector:
+    matchLabels:
+      app: chatbot-rag-app
+  template:
+    metadata:
+      labels:
+        app: chatbot-rag-app
+    spec:
+      # Simplify networking so that it can look for elasticsearch on localhost
+      hostNetwork: true
```
codefromthecrypt, line 17: @anuraaga made a good point offline that if just trying to get things listening on localhost, k8s is the hardest option. Maybe focus this on normal k8s instead, even if it is more work than this PR stands so far.

* instructions to setup Elastic Stack in k8s, including APM server (maybe another directory in this repo like our "docker" one.
* Have to edit the properties in .env to make sure you use the right hostname for elasticsearch and the otel endpoint as localhost won't work anymore

Might be best to go ahead and do that I think, so will unless someone feels otherwise.

codefromthecrypt:
I have work almost done to make this "normal k8s" local, but wanted to solve the timeout first. so I'll push commit after https://github.com/elastic/elasticsearch-labs/pull/397 is merged

codefromthecrypt:
rebased and changed to non-host network k8s. will leave this in draft until https://github.com/elastic/elasticsearch-labs/pull/397 is merged as using not-yet-deployed images in k8s is a pain.

codefromthecrypt:
waiting to get the docker image smaller before "ready for review", as I noticed my network lagging https://github.com/elastic/elasticsearch-labs/pull/407

codefromthecrypt:
ok things work in general, but I'm not seeing traces in kibana. I have to put this down for a bit as I have other more urgent things to address.

k8s/README.md
```diff
@@ -0,0 +1,47 @@
+# Running your own Elastic Stack with Kubernetes
+
+If you'd like to start Elastic with Kubernetes, you can use the provided
+[manifest-elastic.yml](manifest-elastic.yml) file. This starts
+Elasticsearch, Kibana, and APM Server in an existing Kubernetes cluster.
+
+Note: If you haven't checked out this repository, all you need is one file:
+```bash
+wget https://raw.githubusercontent.com/elastic/elasticsearch-labs/refs/heads/main/docker/docker-compose-elastic.yml
```
anuraaga, line 9: Think this is wrong file

k8s/k8s-manifest-elastic.yml
```diff
@@ -0,0 +1,202 @@
...
+              echo "Waiting for kibana to be available";
+              until curl -s http://kibana:5601/api/status | grep -q 'available'; do sleep 5; done;
+              echo "Found kibana";
+      containers:
+        - name: apm-server
+          image: docker.elastic.co/apm/apm-server:8.17.2
+          ports:
+            - containerPort: 8200
+          command:
+          - /usr/local/bin/docker-entrypoint
```
anuraaga, line 182: TBH didn't go so far as running chatbot yet but could confirm traces for normal openai examples with this change

example-apps/chatbot-rag-app/k8s-manifest.yml
```diff
@@ -0,0 +1,57 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: chatbot-rag-app
+  template:
+    metadata:
+      labels:
+        app: chatbot-rag-app
+    spec:
+      # The below will recreate your secret based on the gcloud credentials file
```
anuraaga, line 16: I personally find the GCP-specific instructions in the readme and manifest to be sort of jarring and one point is this is mostly for running locally, while in a cluster the pod should use service account auth provided by the cluster. Similar can be said about using static AWS credentials via env.

"Note for cloud provided backends such as Vertex AI and Bedrock, the pod will need to be authenticated with access to them. This is usually with service account authentication in running clusters, while if testing locally can be with AWS static credentials passed by environment variables or copying a GCP credentials file into a secret mounted as a volume."

Not sure if this is too much information though but I think it would be good if this yaml could avoid the gcloud-specific parts.

codefromthecrypt: good point, I also was troubled by it. better to leave this out as it is arduous and creates FUD.

k8s/README.md
codefromthecrypt: yep

k8s/k8s-manifest-elastic.yml
codefromthecrypt: good eye!

codefromthecrypt:
Due to elasticon singapore and Sydney... while excited about this i am not finishing it this weekend. Maybe Tuesday

example-apps/chatbot-rag-app/k8s-manifest.yml
```diff
@@ -0,0 +1,57 @@
...
+          - name: FLASK_APP
+            value: api/app.py
+        # This recreates your configmap based on your .env file:
+        # kubectl create configmap chatbot-rag-app-env --from-env-file=.env
+        envFrom: &envFrom
+          - configMapRef:
+              name: chatbot-rag-app-env
+        volumeMounts: &volumeMounts
+          - name: gcloud-credentials
+            mountPath: /root/.config/application_default_credentials.json
```
davidgeorgehope, line 37: @codefromthecrypt 
this is not sufficient i was getting errors finding the json. I had to add:

subPath: application_default_credentials.json

The subPath parameter is crucial here - it tells Kubernetes to mount only the specific file from the Secret, not create a directory structure.

davidgeorgehope: mountPath probably should be this too:

mountPath: /root/.config/

bshetti, line 37: SO @davidgeorgehope tested this and you need s subpath and not just a mountpath.

davidgeorgehope: @bshetti here 👍

example-apps/chatbot-rag-app/k8s-manifest.yml
```diff
@@ -0,0 +1,57 @@
...
+      #  --from-file=application_default_credentials.json=$HOME/.config/gcloud/application_default_credentials.json
+      volumes:
+        - name: gcloud-credentials
+          secret:
+            secretName: gcloud-credentials
+      initContainers:
+      - name: create-index
+        image: &image ghcr.io/elastic/elasticsearch-labs/chatbot-rag-app:latest
+        args: ["flask", "create-index"]
+        env:
```
bshetti, line 27: Assuming this walks through the initi also or is there another initi manifest to load up the index?

example-apps/chatbot-rag-app/k8s-manifest.yml
bshetti: I have used static AWS keys with cloud.elastic.co - no service account needed.

example-apps/chatbot-rag-app/env.example
```diff
@@ -68,7 +70,9 @@ OTEL_SDK_DISABLED=true
 # Assign the service name that shows up in Kibana
 OTEL_SERVICE_NAME=chatbot-rag-app
 
-# Default to send traces to the Elastic APM server
+# Default to send logs, traces and metrics to an Elastic APM server accessible
+# via localhost. If using running inside k8s, update to http://apm-server:8200
+# or similar.
```
bshetti, line 75: We need to add a section for running in cloud.elastic.co

example-apps/chatbot-rag-app/k8s-manifest.yml
bshetti: Agreed. I just tried it also, you need to add subpath also

codefromthecrypt: cool, so @anuraaga had mentioned that due to complexity of this and also that not all will even use vertex, should we continue to make configurations per-cloud by default, or instead delete this section and document you should edit it. thoughts? cc @joemcelroy as this is one of the things that while I do think we can sort out the glitch here easy enough due to folks pointing it out.. we do need to ask ourselves if we are digging a hole, or if we are doing the right thing. Basically should we bake in things about clouds that are complicated and would be kindof obnoxious to read about if you don't care about vertex?

example-apps/chatbot-rag-app/k8s-manifest.yml
codefromthecrypt: me too. I think the main point is the stuff needed to get GCP working which is easier to hide in docker compose than k8s

example-apps/chatbot-rag-app/k8s-manifest.yml
bshetti: I don't think it is a bad idea to leave it. Not having Vertex will look weird. Since it requires the JSON and we require this to make it work its an option. I can try to fix the README. Let me give it a go tonite or tomorrow.

codefromthecrypt: Actually slept on it and I think we can keep it and also avoid README scare by making the instructions collapsable! Will give a go and we can see how it looks.

codefromthecrypt: ok I think last commit sorts this. testing local and thanks @davidgeorgehope and @bshetti for the config tips

codefromthecrypt:
hmm getting gcp auth errors will look into it

example-apps/chatbot-rag-app/k8s-manifest.yml
codefromthecrypt: actually, it if created by the readme, mountPath is needed, at the dir, but not subPath as the filename is already named application_default_credentials.json when creating the secret.

example-apps/chatbot-rag-app/k8s-manifest.yml
codefromthecrypt: yes this is same as docker, creates everything.

example-apps/chatbot-rag-app/env.example
codefromthecrypt: please do elastic cloud stuff in a follow-up as these instructions also need to apply to python and docker. I opened https://github.com/elastic/elasticsearch-labs/issues/379 for this topic. this PR is hairy enough with the changes it has so far.

codefromthecrypt:
GCP vertex now works. I will look into why traces aren't.

@bshetti I can't hold this PR captive for all issues, as once this is in it is easy to complete other topics. So, let's leave elastic cloud commentary for the next PR https://github.com/elastic/elasticsearch-labs/issues/379 This one is solving as-is for k8s, and it has been dozens of hours just on that!

codefromthecrypt:
Also verified the kubernetes without chatbot-rag-app, rather with pydantic-ai and works fine

Screenshot 2025-03-02 at 11 19 51 AM

codefromthecrypt:
create-index (doesn't use an LLM, just elastic)

Screenshot 2025-03-02 at 11 24 05 AM

chat (proves vertex works)

Screenshot 2025-03-02 at 11 25 12 AM

codefromthecrypt:
in this case I followed the directions in the README with a completely blown away k8s (`colima delete; colima start --cpu 8 --memory 16 --network-address --dns 8.8.8.8 --dns 8.8.4.4 --kubernetes --k3s-arg '--disable=local-storage,traefik,metrics-server@server:*'`), so I'm very confident the GCP stuff works as nothing was dirty. Thanks for the tips, folks!

example-apps/chatbot-rag-app/k8s-manifest.yml
```diff
@@ -0,0 +1,55 @@
...
+  template:
+    metadata:
+      labels:
+        app: chatbot-rag-app
+    spec:
+      # For `LLM_TYPE=vertex`: create a volume for application_default_credentials.json
+      volumes:
+        - name: gcloud-credentials
+          secret:
+            secretName: gcloud-credentials
```
anuraaga, line 20: Sorry realized one thing post approve - how about populating the secret in this file with a dummy value? I think currently the deployment will fail if the step to create the secret manually isn't done.

codefromthecrypt: lemme go from scratch not creating the secret. if true I'll try to figure out how to do what you've mentioned as indeed we don't want a failure on something optional and for many obscure.

codefromthecrypt: totally right..

```
Events:
  Type     Reason       Age               From               Message
  ----     ------       ----              ----               -------
  Normal   Scheduled    39s               default-scheduler  Successfully assigned default/chatbot-rag-app-7984995bc8-pwqxg to colima
  Warning  FailedMount  8s (x7 over 40s)  kubelet            MountVolume.SetUp failed for volume "gcloud-credentials" : secret "gcloud-credentials" not found
```

codefromthecrypt: got that working, then ran it by grok 3 think mode which told me all I need to do is mark this optional ;)

codefromthecrypt:
OK, what I did was ran with the normal instructions, but azure openai (so no secret). It worked fine.
Screenshot 2025-03-02 at 2 03 44 PM

Then, I deleted the configmap and edited in the vertex settings to recreate it, then added the secret as README said, then applied and worked fine.

Screenshot 2025-03-02 at 2 05 54 PM

Thanks for the eagle eyes @anuraaga I think finally this one is ready to merge!

example-apps/chatbot-rag-app/k8s-manifest.yml
```diff
@@ -0,0 +1,58 @@
...
+            optional: true  # only read when `LLM_TYPE=vertex`
```
codefromthecrypt, line 23: this part allows vertex config to work, but others to not block on it. the optional applies indirectly to a mount that uses it, so no worries.

PR #406, opened 2025-02-25
chatbot-rag-app: removes unnecessary parts of Docker build

codefromthecrypt:
This removes the temporary workaround for the OTEL_SDK_DISABLED variable


codefromthecrypt:
moved the parts not about the entrypoint to this PR meanwhile https://github.com/elastic/elasticsearch-labs/pull/407

codefromthecrypt:
when https://github.com/elastic/elastic-otel-python-instrumentations/pull/54 is released, we should be good to go here

codefromthecrypt:
ok I tried this both ways, enabled and disabled. no worries.

I can't try k8s because the docker image is pushed only after this change. However, it should work.

anuraaga (approved):

PR #407, opened 2025-02-25
chatbot-rag-app: reduces docker image size

codefromthecrypt:
This reduces the chatbot-rag-app image size from 1.3GB to 820MB by eliminating ancillary packages.

This is more needed when we use k8s which doesn't share an image cache with docker, typically. When demo'ing we want the least download lag possible. Also, my network is not always very strong, so prefer smaller.

While at it, I did a little polishing, too.

Anecdotally, I tried to switch to alpine, but the amount of system packages needed resulted in a larger image than it is now!

### DIFF
--- a/docker/docker-compose-elastic.yml
+++ b/docker/docker-compose-elastic.yml
@@ -69,7 +69,7 @@ services:
       - XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY=fhjskloppd678ehkdfdlliverpoolfcr
       - SERVER_PUBLICBASEURL=http://127.0.0.1:5601
     healthcheck:
-      test: ["CMD-SHELL", "curl -s http://localhost:5601/api/status | grep -q 'All services are available'"]
+      test: ["CMD-SHELL", "curl -s http://localhost:5601/api/status | grep -q 'available'"]
       retries: 300
       interval: 1s
 
--- a/example-apps/chatbot-rag-app/Dockerfile
+++ b/example-apps/chatbot-rag-app/Dockerfile
@@ -5,24 +5,19 @@ COPY frontend ./frontend
 RUN cd frontend && yarn install
 RUN cd frontend && REACT_APP_API_HOST=/api yarn build
 
+# langchain and vertexai depend on a large number of system packages including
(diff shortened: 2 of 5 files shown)


docker/docker-compose-elastic.yml
```diff
@@ -69,7 +69,7 @@ services:
       - XPACK_ENCRYPTEDSAVEDOBJECTS_ENCRYPTIONKEY=fhjskloppd678ehkdfdlliverpoolfcr
       - SERVER_PUBLICBASEURL=http://127.0.0.1:5601
     healthcheck:
-      test: ["CMD-SHELL", "curl -s http://localhost:5601/api/status | grep -q 'All services are available'"]
+      test: ["CMD-SHELL", "curl -s http://localhost:5601/api/status | grep -q 'available'"]
```
codefromthecrypt, line 72: easier match, so keep it simple

example-apps/chatbot-rag-app/README.md
```diff
@@ -165,8 +165,6 @@ pip-compile
 pip install -r requirements.txt
 # Add opentelemetry instrumentation for these dependencies
 edot-bootstrap >> requirements.txt
-# Missing dependency for langtrace vertexai instrumentation
```
codefromthecrypt, line 168: fixed in langtrace 3.3.28

example-apps/chatbot-rag-app/Dockerfile
```diff
@@ -5,24 +5,19 @@ COPY frontend ./frontend
...
+# on -slim and adding them to -alpine results in a larger image than -slim.
 FROM python:3.12-slim
 
 WORKDIR /app
 RUN mkdir -p ./frontend/build
 COPY --from=build-step ./app/frontend/build ./frontend/build
-RUN mkdir ./api
-RUN mkdir ./data
-
-RUN apt-get update && apt-get install -y \
```
codefromthecrypt, line 16: none of these are required

codefromthecrypt:
cc @anuraaga

codefromthecrypt (commented):
notes

anuraaga (approved):

PR #410, opened 2025-02-25
Bugfix in Crawler Migration Notebook extraction rule cell

mattnowzari:
### Closes https://github.com/elastic/search-team/issues/9376

This PR fixes a bug was found in the Elastic Crawler notebook where the extraction rule cell was attempting to add extraction rules for domains that technically no longer exist. This is due to how certain actastic indices work (the extraction rules index doesn't seem to be kept in sync with the domains index)

This PR also fixes a related bug where a Crawler with multiple extraction rules was only receiving the last-known extraction rule instead of _all_ extraction rules


notebooks/enterprise-search/elastic-crawler-to-open-crawler-migration.ipynb
```diff
@@ -311,13 +311,15 @@
...
     "    ]\n",
     "\n",
-    "    print(\n",
-    "        f\"{extr_count}.) Crawler {config_oid} has extraction rules {temp_extraction_rulesets}\\n\"\n",
-    "    )\n",
-    "    extr_count += 1\n",
-    "    inflight_configuration_data[config_oid][\"domains_temp\"][domain_oid][\n",
-    "        \"extraction_rulesets\"\n",
-    "    ] = temp_extraction_rulesets"
+    "    if domain_oid in inflight_configuration_data[config_oid][\"domains_temp\"]:\n",
```
codefromthecrypt, line 314: maybe add a comment above this, considering if you did the steps prior why the guard is needed?

mattnowzari: WIll do!
In addition, I might just move this check further up anyways because we get `domain_oid` early enough that we can avoid a lot of steps altogether if we know the domain doesn't actually exist

JessicaGarson (approved):

PR #458, opened 2025-05-27
chatbot-rag-app: updates to latest openai and uses uv to compile deps

codefromthecrypt:
The published image was missing the chat span, this sorts it out. This also switches to uv to bump deps because pip compile was hanging due to complexity.

Screenshot 2025-05-27 at 10 50 23 AM


example-apps/chatbot-rag-app/README.md
```diff
@@ -182,7 +179,7 @@ dotenv run -- flask create-index
 
 Now, run the app, which listens on http://localhost:4000
 ```bash
-dotenv run -- python api/app.py
+uv run -q --env-file .env python api/app.py
```
SrdjanLL, line 182: I would suggest to also update the OpenTelemetry section below in the doc to replace all commands using `dotenv` the `uv` to make the docs consistent with this change.

PR #471, opened 2025-07-15
DRAFT: Added an updated version of notebook for video script

justincastilla:
Updated the notebook to use modern mechanisms and API calls. I ONLY updated the notebook codebase, not the supporting text. I will change the copy at a later time


carlyrichmond (changes requested):
I like the summary being added. Can you make sure all TODO's are addressed and the comments removed? After that I think we're good.

As discussed on Wednesday, let's also make sure this updated notebook replaces the original once the piece has been updated.

carlyrichmond (changes requested):
I think there's still 1 TODO comment in there, but aside from that LGTM.

PR #476, opened 2025-07-31
Fix LTR notebook in Google Collab

afoucret:
The `q` param prevent Google collab to detect change in numpy and prompt for a restart. The user experience feels broken.


markjhoy (approved):
👍

PR #486, opened 2025-09-10
Blog content for Twelvelabs + Bedrock + Elasticsearch

derickson:
In support of blog post with Twelvelabs.io
Notebook tested, outputs saved into notebook.  should be in an isolated folder.


derickson:
Found a bug that needs a bigger change  cancelling this PR

PR #491, opened 2025-09-29
Breaking Language Barriers with Cross-Lingual Vector Search Python Notebook

qn895:


carlyrichmond:
Thanks for adding the code in a notebook @qn895! I've tweaked the connection settings to use API Key and endpoint instead of cloud id and the basic auth without credentials to show best connection practice.

Could you also add an example query to the end of the notebook as well from the piece for completeness? Once that's in it should be good to go in my view. 

Hope that helps!

qn895:
@carlyrichmond Just updated the notebook with more example queries, thank you for the feedback!

PR #507, opened 2025-11-11
Justin black jupyter test

justincastilla:
This PR bumps our pre-commit/actions to use the latest version of black-jupyter (25.11.0)

Version 24.1.1 is breaking during the pre-commit/actions black-jupyter check on notebooks. This is patched with pinning the Python version that uses black-jupyter @3.13, but best practice would be to update black-jupyter, as our global version is 3.14.

Made a small change to a notebook to test the pre-commit.


justincastilla:
@carlyrichmond this is the PR to update jupyter-black. The pre-commit tests in this PR are showing success w/ the version bump.

PR #548, opened 2026-04-30
supporting blog content google-adk-voice-assistant-with-agent-builder

Delacrobix:


someshwaranM (changes requested):
Hi @Delacrobix,

One suggestion: if we skip the README, could we at least include this context in the top section of this notebook? For example, we can mention that this notebook covers the Elastic portion of the workflow, while the Google ADK-related codebase and end-to-end context are covered in the blog. Attached a screenshot below:

Image

This would help developers who land directly in the repo understand the notebook’s scope, what it covers, and where to go for the ADK-related details. It also creates a clear connection between the notebook and the blog, so users can move back and forth regardless of where they start.

Delacrobix:
Hi @someshwaranM 

Thank you so much for your suggestion! Sure, I just added that at the beginning of the notebook.

someshwaranM (approved):
This looks good to me. Thanks.

